PR #42, opened 2023-11-14
chore: move to nx monorepo integrated mode

alvaroraminelli:
**What changed? Why?**

Move to nx integrated approach by implementing shared eslint, tsconfig and prettier. It will be easy to add brand-new project because all the tooling decision have already been made. Trafe-off: Hard to add existing projects because the build tooling for that package may need to be modified.

Changes:
- shared eslint, prettier and tsconfig
- verdaccio for local npm registry for help testing local cli

## Local Development

To test the CLI locally, run:

1. Run the local NPM registry:

```
npx nx local-registry
```

2. In a different window, publish a new version of the packages:

```
npx nx run-many -t publish --ver 0.0.1 --tag latest
```

**Notes to reviewers**

**How has it been tested?**


Zizzamia (approved):

Zizzamia:
Btw, more a curiosity, I understand we don't want "Package-Based", but why "Integrated" vs "Standalone Apps".

PR #46, opened 2023-11-15
Enable npm provenance

wespickett:
**What changed? Why?**
Enable npm provenance

**Notes to reviewers**

**How has it been tested?**
Confirmed it working here:
https://www.npmjs.com/package/npm-provenance-test2?activeTab=versions

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -2,15 +2,15 @@ name: Setup
 runs:
   using: composite
   steps:
-    - uses: pnpm/action-setup@v2
-      with:
-        version: 8.6.0
-
     - uses: actions/setup-node@v3
       with:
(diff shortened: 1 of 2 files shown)


.github/actions/setup/action.yml
```diff
@@ -2,16 +2,16 @@ name: Setup
 runs:
   using: composite
   steps:
-    - uses: pnpm/action-setup@v2
-      with:
-        version: 8.6.0
-
```
wespickett, line 8: @Zizzamia It doesn't seem like this is needed?

Zizzamia: Yeah your right, great catch!

Zizzamia (approved):
That's brilliant, thank you so much.

PR #107, opened 2023-11-29
feat: Update Navbar styles and move it into the Header

codynova:
**What changed? Why?**

* Update Navbar styles and move it into the Header

**Notes to reviewers**

**How has it been tested?**

* Local review


Zizzamia (approved):
Niiiice

PR #119, opened 2023-12-02
feat: Add foundry support

Sneh1999:
**What changed? Why?**
- Added support for Foundry for BuyMeACoffee.sol
- Added documentation, tests, deployment, and verification scripts

**Notes to reviewers**

**How has it been tested?**
Unit tests


apps/build-onchain-apps/contracts/.env.example
```diff
@@ -0,0 +1,3 @@
+PRIVATE_KEY=
+BASE_GOERLI_RPC_URL="https://goerli.base.org"
```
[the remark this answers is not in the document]
Sneh1999: How it works is as follows: It reads from `.env` I believe
```
$ source .env
$ forge script script/BuyMeACoffee.s.sol:BuyMeACoffeeScript --broadcast --verify --rpc-url ${BASE_GOERLI_RPC_URL} --etherscan-api-key ${BASESCAN_API_KEY}
```

Added the instructions in the README.md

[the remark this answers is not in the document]
Sneh1999: What do you suggest? Should we just modify the command line script to the hard-coded arguments?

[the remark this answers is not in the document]
Sneh1999: okay I will update this!

apps/build-onchain-apps/contracts/README.md
```diff
@@ -0,0 +1,75 @@
...
+$ forge fmt
+```
+
+### Deploy to Base Goerli
+
+Create a `.env` file using the `.env.example` file provided.
+
+```shell
+$ source .env
+$ forge script script/BuyMeACoffee.s.sol:BuyMeACoffeeScript --broadcast --verify --rpc-url https://goerli.base.org --etherscan-api-key ${BASESCAN_API_KEY}
```
cnasc, line 64: @Sneh1999 Ah I even would have just hardcoded it in the toml file

cnasc: Yeah so maybe just `RPC_URL` and also `BLOCK_EXPLORER_URL` so that we don't make the variable names network-specific? Although I guess then we run into having to make the object keys generic as well

Sneh1999: I will update the names for now. I think this is a starting point but yes, as we progress we can make things more general.

apps/build-onchain-apps/contracts/foundry.toml
```diff
@@ -0,0 +1,13 @@
+[profile.default]
+src = "src"
+out = "out"
+libs = ["lib"]
+fs_permissions = [{ access = "read", path = "./"}]
+
+[rpc_endpoints]
+base_goerli="${RPC_URl}"
```
cnasc, line 8: I'm not sure if something like this works in toml, but it seems like for this be a full solution we'd need the network name also to be dynamic. Otherwise we end up in the same place where if it's specific to a network we might as well hardcode the value for that. 

```suggestion
${NETWORK}="${RPC_URl}"
```

Sneh1999: Seems like it wont work
image

Sneh1999: Makes sense! I tried it and it works :) Updated it now

PR #128, opened 2023-12-05
feat: setting up the initial Jest configuration

Zizzamia:
**What changed? Why?**
Setting up the initial Jest configuration so we can start writing test for TS and TSX files.

**Notes to reviewers**

**How has it been tested?**


Zizzamia:
Tested this with @wespickett, we are going to keep polishing on this in the next PR.

Adaptcontractor:
382247

PR #153, opened 2023-12-12
fix: move to different app structure

wespickett:
**What changed? Why?**
Moved pages to hold only pages, and components/hooks/styles under src
Removed ts paths since they tend to cause issues

**Notes to reviewers**

**How has it been tested?**
yarn build
yarn dev


Zizzamia (approved):
Thank you for this fix!

PR #154, opened 2023-12-13
Restructure contracts folder

Sneh1999:
**What changed? Why?**
- Initializing git on create for the user

**Notes to reviewers**

**How has it been tested?**


apps/build-onchain-apps/contracts/.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
+name: test
+
+on: workflow_dispatch
+
+env:
+  FOUNDRY_PROFILE: ci
+
+jobs:
```
Zizzamia, line 8: Wait, we can have `.github` files in the different part of the app, that's neat!!
Ok I am going to move the other as well ✨

Zizzamia: Maybe better question, would it be better to have this file as `apps/build-onchain-apps/.github/~~~`???
So all Workflow related to `apps/build-onchain-apps` are all in the same place?

src/utils/git.ts
```diff
@@ -0,0 +1,24 @@
+import { execSync } from 'child_process';
+
```
Zizzamia, line 2: That's a power move, love this!

apps/build-onchain-apps/contracts/.github/workflows/test.yml
Zizzamia: Ok, I actually give it a try, and it seems can only be one `.github` and collection of workflows at the root of the project. So I think we need to move this to the way top to have it working.

Sneh1999: Thank you! For pointing this out, I will fix this. 

It's still a draft PR though, I still need to get some stuff working, I will let you know once it's done and feel free to review it again then 😄

PR #159, opened 2023-12-13
chore: add stylelint

wespickett:
**What changed? Why?**
Added stylelint checks, and sort order "stylelint-config-idiomatic-order"

**Notes to reviewers**

**How has it been tested?**
Added as a CI check


Zizzamia (approved):
Niiice

PR #168, opened 2023-12-15
[chore] Restructure contracts folder

Sneh1999:
**What changed? Why?**
- Initializes git for the user
- Reduces the number of commands to run for the user in the contracts folder

**Notes to reviewers**

**How has it been tested?**
local testing


.github/workflows/foundry.yml
```diff
@@ -0,0 +1,34 @@
+name: test
```
Zizzamia, line 1: Interesting, kind of curios to see why this workflow is not running on this PR 🤔

Zizzamia (approved):
Love it!!!

.github/workflows/foundry.yml
Zizzamia: Oh I see was missing this
```
on:
  push:
    branches: ['main']
  pull_request:
    branches: ['main']
```

PR #180, opened 2023-12-19
Adding support for API routes in the NextJS layer + Configuration

robpolak-cb:
**What changed? Why?**
1. **Configuration / Environment Settings** - Adds a new configuration construct where we can start feeding in environment parameters in a cohesive manner.
2. **Backend API Routes** - Added an example Next.js API route.  (More below on why I chose to use Next.js over more seasoned frameworks like ExpressJs)
3. **Chain Configuration** - Added a function to get the supported chains for an environment.  This was needed because we will need to chain ID on the API backend for lookups.  This will also be needed so we can start mapping other configuration to a specific environment.  For example, we might want to support BuyMeACoffee on both Goerli and Sepolia testnets.  Additionally, an end user might want to support multiple EVMs for their project (so we should support that).

**Notes to reviewers**
1. **Why Next.JS routes over Express JS** - **_TLDR_**; Keeping it simple until we have a need to demo a complex API implementation.
   2.  Running Express inside the NextJS container is not optimal - There are stylistic and dependency differences between Next + Express.  While you can get both running together you end up having to compromise on some pretty ugly convention (e.g. CustomServer).  Likely if we wanted to do this the right way, we should be spinning up a totally new project which only has an API endpoint.
   3. Keep it simple - While it is common some folks need a robust API layer, most small projects can live with the limitations of running within Next API routes. 

**How has it been tested?**
1. next dev + verified locally
4. yarn test + wrote additional test for new code.


apps/build-onchain-apps/pages/api/README.md
```diff
@@ -0,0 +1,11 @@
+# Build Onchain API Routes
+
+## Supported Chains /api/chains/supported
+
+**Method**: GET
+
+**Response**: JSON encoded viem chain object
+
+```json
+[{"id":5,"network":"goerli","name":"Goerli",...]
```
Sneh1999, line 10: we should change this to base-goerli maybe? Thoughts? rest of the app currently uses that as an example

apps/build-onchain-apps/pages/api/chains/supported.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    (getCurrentEnvironment as jest.Mock).mockReturnValue('development');
+    (getChainsForEnvironment as jest.Mock).mockReturnValue(['chain1', 'chain2']);
+
+    await handler(req as NextApiRequest, res as NextApiResponse);
+
+    expect(mockStatus).toHaveBeenCalledWith(200);
+    expect(mockJson).toHaveBeenCalledWith(['chain1', 'chain2']);
+  });
+
+  it('handles errors gracefully', async () => {
```
Sneh1999, line 40: Beautiful!

apps/build-onchain-apps/pages/api/chains/supported.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param req
+ * @param res
+ */
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    const chains = getChainsForEnvironment();
+    res.status(200).json(chains);
+  } catch (error) {
+    console.error('Error fetching chains:', error);
+    res.status(500).json({ error: 'Internal server error' });
```
Sneh1999, line 16: Should we pass the error as well, in the response?

apps/build-onchain-apps/pages/api/README.md
Sneh1999: Also maybe we should also add this [here](https://github.com/coinbase/build-onchain-apps/blob/main/README.md)?

Sneh1999 (commented):
Looks great! Thank you

apps/build-onchain-apps/pages/api/chains/supported.ts
robpolak-cb: Typically you don't want to pass along errors in production use-cases (as it could expose an attack surface, e.g. database error).  I do think it is reasonable to pass the error along in dev/localhost

apps/build-onchain-apps/pages/api/README.md
robpolak-cb: It is base-goerli :) see chainConfiguration.ts this is just an example payload.  I didn't want to clutter the main readme.

Sneh1999 (approved):

apps/build-onchain-apps/src/utils/chainConfiguration.ts
```diff
@@ -0,0 +1,22 @@
+import { baseGoerli, Chain, goerli } from 'viem/chains';
+import { Environment } from './environment';
+import { getCurrentEnvironment } from './configuration';
+
+// The list of supported Chains for a given environment (e.g. Dev should only have testnet)
+export const supportedChains = new Map<Environment, Chain[]>([
+  [Environment.localhost, [baseGoerli]],
+  [Environment.development, [baseGoerli]],
+  [Environment.staging, [goerli]],
+  [Environment.production, [goerli]],
```
Zizzamia, line 10: nit. we want to be the default always being `base`, than others are optional choices.

PR #185, opened 2023-12-20
Adding Block Number API + React Hook

robpolak:
**What changed? Why?**
Added a blockNumber api + react hook for the FE to consume from the BE API.

1. Added new API route  `/api/chain/currentBlockNumber?chainId=${chainId}`
2. Added a new reactHook useCurrentBlockNumber
   1. Polls for the block number every 10 seconds by default
   2. Added to the AccountConnectButton feature.

**Notes to reviewers**
1. Why would we want blockNumber on the backend?

Sometimes it is helpful to get the block number from the backend to ensure the FE is in sync with network.  And mainly, that our FE And BE are in alignment.  Not the most critical feature, but this showcases how we would make a call to the BE from a react hook

2. Is interval the best idea in a react hook?

Probably not, we should implement a global state handler, will let some of react pros tackle this one @cnasc @snehkoul-cb 

**UX Changes**
Added network + blockNumber to the Account dialog: 
image

**How has it been tested?**
1. Unit test + ran locally

### DIFF
--- a/apps/build-onchain-apps/src/hooks/useBlockNumber.ts
+++ b/apps/build-onchain-apps/src/hooks/useBlockNumber.ts
@@ -27,7 +27,7 @@ const useCurrentBlockNumber = (refreshIntervalMs = 10000) => {
           throw new Error(`Error: ${response.status}`);
         }
 
-        const data: BlockNumberResponse = (await response.json()) as BlockNumberResponse;
+        const data = (await response.json()) as BlockNumberResponse;
         setBlockNumber(data.block);
       } catch (err) {
         // Handle error
--- a/apps/build-onchain-apps/src/hooks/useBlockNumber.ts
+++ b/apps/build-onchain-apps/src/hooks/useBlockNumber.ts
@@ -1,0 +1,55 @@
+import { useState, useEffect } from 'react';
+import { useNetwork } from 'wagmi';
+
+type BlockNumberResponse = {
+  block: number;
+};
+const useCurrentBlockNumber = (refreshIntervalMs = 10000) => {
+  const [blockNumber, setBlockNumber] = useState(0);
+  const [isLoading, setIsLoading] = useState(true);
+  const { chain } = useNetwork();
+  const chainId = chain?.id;
+
+  useEffect(() => {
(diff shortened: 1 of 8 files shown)


apps/build-onchain-apps/src/hooks/useBlockNumber.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Refresh current block number every 15 seconds.
+    const fetchBlockNumberWrapper = () => {
+      fetchBlockNumber().catch(console.error); // Handle promise rejection
+    };
+    // TODO: Implement a shared state or other mechanism to prevent overloading the backend with multiple intervals.
+    intervalId = setInterval(fetchBlockNumberWrapper, refreshIntervalMs);
+    void fetchBlockNumber();
+
+    return () => {
+      clearInterval(intervalId); // Clear interval on cleanup
```
cnasc, line 48: 👍 `setInterval` in a hook is pretty idiomatic, the only hitch is to return a cleanup function which you have done so A+

apps/build-onchain-apps/src/hooks/useBlockNumber.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      setIsLoading(true);
+
+      try {
+        const response = await fetch(`/api/chain/currentBlockNumber?chainId=${chainId}`);
+        if (!response.ok) {
+          throw new Error(`Error: ${response.status}`);
+        }
+
+        const data: BlockNumberResponse = (await response.json()) as BlockNumberResponse;
```
cnasc, line 30:
```suggestion
        const data = (await response.json()) as BlockNumberResponse;
```
minor nit, but this should be inferred due to the cast

cnasc (approved):
LGTM, you could use https://react.dev/reference/react/createContext to manage this so the interval is a singleton and consumers can read from the context state

apps/build-onchain-apps/src/hooks/useBlockNumber.ts
robpolak: fixed

Zizzamia (approved):

PR #186, opened 2023-12-20
feat: basic multichain config

cnasc:
**What changed? Why?**
Adding some types and a config file that will make it easier for us to support multiple chains. Pushing this up before any further implementation to get feedback on the approach taken here.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/apps/build-onchain-apps/src/contract/ContractSpecification.ts
+++ b/apps/build-onchain-apps/src/contract/ContractSpecification.ts
@@ -1,0 +1,26 @@
+import { Abi, type Chain } from 'viem';
+import { baseGoerli } from 'viem/chains';
+import BuyMeACoffeeABI from './BuyMeACoffee';
+
+export type ContractInstance = {
+  chain: Chain;
+  address: `0x${string}`,
+  deactivated?: boolean,
+}
+
+export type ContractSpecification = Record<string, {
+  abi: Abi;
+  [chainId: number]: ContractInstance;
+}>;
+
+export const contracts = {
+  buyMeACoffee: {
+    abi: BuyMeACoffeeABI,
+    [baseGoerli.id]: {
+      chain: baseGoerli,
+      address: '0xeeC0bD3B58293ff45004C7eFf02917Beef28257c',
+    }
+    // more chains for this contract go here
+  }
+  // more contracts go here
+}


cnasc:
@robpolak how does this look?

apps/build-onchain-apps/src/contract/ContractSpecification.ts
```diff
@@ -0,0 +1,26 @@
+import { Abi, type Chain } from 'viem';
+import { baseGoerli } from 'viem/chains';
+import BuyMeACoffeeABI from './BuyMeACoffee';
+
+export type ContractInstance = {
+  chain: Chain;
+  address: `0x${string}`,
```
robpolak, line 7: What is this black magic

robpolak (approved):
LGTM

apps/build-onchain-apps/src/contract/ContractSpecification.ts
cnasc: Basically “a string that starts with `”0x”`. Compiler can only tell from string literals naturally

Zizzamia: Harry Potter vibes

apps/build-onchain-apps/src/contract/ContractSpecification.ts
```diff
@@ -0,0 +1,26 @@
+import { Abi, type Chain } from 'viem';
+import { baseGoerli } from 'viem/chains';
+import BuyMeACoffeeABI from './BuyMeACoffee';
+
+export type ContractInstance = {
```
Zizzamia, line 5: if we are exporting `types`, should at this point have just have a `types.ts` file in this folder that we use to organize different kind of Contract types we need?

cnasc: I usually prefer to have the types colocated with any implementation code, but open to your approach as well. Do you have a strong preference?

Zizzamia (approved):

apps/build-onchain-apps/src/contract/ContractSpecification.ts
Zizzamia: Approved and merged the PR as this is not a blocker.

My thinking is about where and how things can be re-use. I don't want to jump on conclusion, so happy to just revisit this as we have more code to read :-)

LFG

PR #193, opened 2023-12-23
feat: basic multichain support

cnasc:
**What changed? Why?**
Replaced the old contract specification object with hooks that communicate network connection status. Additionally, updated our ABIs to allow for better types through wagmi. There is some more work to be done here (unified network switching component that takes a list of supported networks, etc.) in the future.

**Notes to reviewers**

**How has it been tested?**
Added unit tests for the hook generator, manual testing.


apps/build-onchain-apps/src/contract/BuyMeACoffee.ts
```diff
@@ -114,6 +112,6 @@ const abi: Abi = [
     type: 'function',
   },
   { stateMutability: 'payable', type: 'receive' },
-];
+] as const;
```
cnasc, line 115: See: https://wagmi.sh/react/prepare-hooks/usePrepareContractWrite#abi-optional

We should declare abis `as const` and just let the types be inferred. If we define them as the `Abi` type, it's too broad and we don't get good typechecking. 

With this approach we get nice autocomplete and errors on `functionName`, for example.

image

apps/build-onchain-apps/src/contract/ContractSpecification.ts
```diff
@@ -1,38 +1,69 @@
...
+  | { abi: T; status: 'onUnsupportedNetwork' }
+  | { abi: T; status: 'notConnected' }
+  | { abi: T; status: 'deactivated' };
+
+type Spec<T extends Abi> = {
+  abi: T;
+  [chainId: number]: ContractInstance;
+};
+
+export function generateContractHook<T extends Abi>({ abi, ...spec }: Spec<T>) {
```
cnasc, line 24: generating a hook per contract makes things a little more succinct at the callsite, while not being significantly more verbose than the previous plan of single big object of all contracts with a hook that selects from them.

apps/build-onchain-apps/src/hooks/useOnchainCoffeeMemos.ts
```diff
@@ -12,19 +11,22 @@ import type { CoffeeMemo } from '../types';
...
 function useOnchainCoffeeMemos() {
+  // TODO: this state should be cleared on network change, otherwise memos from multiple chains
+  // could end up in the same list
   const [memos, setMemos] = useState<CoffeeMemo[]>([]);
 
   const handleSuccess = useCallback((newMemos: CoffeeMemo[]) => {
     setMemos(newMemos);
   }, []);
 
+  const contract = useBuyMeACoffee();
```
Zizzamia, line 22: Should we standardize the prefix for those Hooks, like `useContractBuyMeCoffee`?

apps/build-onchain-apps/src/contract/ContractSpecification.ts
Zizzamia: move this hook under `onchain/hooks` folder, as this can be an interesting reusable hook.

apps/build-onchain-apps/src/components/buy-me-coffee/FormBuyCoffee.tsx
```diff
@@ -75,6 +76,18 @@ function FormBuyCoffee({ onComplete }: FormBuyCoffeeProps) {
     [setMessage],
   );
 
+  if (contract.status === 'notConnected') {
```
Zizzamia, line 79: If those are common for all contracts, I wonder if we should expose something that makes this logic even simpler.... I guess that can be a second step. Let's probably explore what's the full experience and than we can simplify even more.

apps/build-onchain-apps/onchainKit/hooks/useBlockExplorerLink.ts
```diff
@@ -11,7 +11,11 @@ export enum HashType {
  * @param {string} hash - The contract or transaction hash.
  * @returns {string} - The URL to the Etherscan page for the given hash.
  */
-function useBlockExplorerLink(chain: Chain, hash: string, type: HashType = HashType.Address) {
+function useBlockExplorerLink(
```
cnasc, line 14: does this really need to be a hook?

Zizzamia: Oh I see your point, should this just be an utility like `getBlockExplorerLink`?

After all if the chain is changing than someone can just re-run the utility and get a fresh `link`.

Right?

@renanmav thoughts on this?

apps/build-onchain-apps/src/onchain/hooks/contracts.test.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
Zizzamia, line 1: Btw, we deprecated the `onchain` folder, so you can just move this into `hooks` directly.

Zizzamia: So in ` apps/build-onchain-apps/src/hooks/contracts.ts`

apps/build-onchain-apps/src/hooks/useCollectionMetadata.ts
```diff
@@ -0,0 +1,68 @@
+import { useCallback, useEffect, useState } from 'react';
```
Zizzamia, line 1: This one, now lives in `apps/build-onchain-apps/onchainKit/`

Zizzamia (approved):
Niiiiice
LFG

PR #213, opened 2024-01-03
Feat: Add Signature Mint Contract + Api Route

robpolak:
**What changed? Why?**

### Contracts changes
1. Added a new contract SignatureMintERC721, which is a varation of a NFT contract which allows FreeMints via a backend wallet.
    1.  Wallet should be defined in the contracts folder via the SIGNATURE_MINT_SIGNER environment param
    2. Includes logic via OpenZepplin to verify a web3 signature.
    3. Has static metadata similar to an open edition (e.g. all tokens have the same metadata)
    4. 
2. Created a NFTUtilities Library for sharing common functions across NFT Contract.
3. Updated readme + added instructions for coverage reporting

### FE/API changes
1. Added a new API route : `/api/mint/signature/free` which will vend a backend signature.
2. Added a new environment parameter : SIGNATURE_MINT_PRIVATE_KEY which will hold the private key.
3. Extracted getChainById to a common function to use across API routes.
4. Fixed linting issues with [CodeBlock.tsx](https://github.com/coinbase/build-onchain-apps/pull/213/files#diff-830e2c3eab6ba574e916c4a2849bd89f8e981af9885792ba5a344953942e8fc0) (I didn't make any changes)

**Open Questions**

## Should we store a sample private key in the protect?
For this to work end-to-end the backend needs a private key.   We could vend this as part of the environment configuration default, however, this is setting a bad example for consuming users (as you should never store a private key in code).  So we have a couple options:
1. Add private key to code - Allows users to pull down a fully functional codebase without having to make any changes (e.g. run the app and free mint).
2. Have the user deploy their own contract - Give the user instructions to deploy their own contract (with their own signer).  This would take a few extra steps and would not work out of the box.

## Discussion Result  (TLDR; Do not store key)
We will not store the private key in code as that is a bad practice we don't want to exhibit.  Rather, we will detect when a key is not set and give the user some extra guidance on the UI.  I will start working on this in my next PR.

**How has it been tested?**
1. Example NFT : https://testnets.opensea.io/assets/base-goerli/0xd68d1a72647de2af10a87db895cb0b8cf477fcd0/1
2. Free Mint Transaction : https://goerli.basescan.org/tx/0xd1429515b727f29acb51b4dd15b23cb981a2700c34d2c19bbd432556ce96a427
3. yarn && yarn build && yarn test && yarn lint :
image
image


apps/build-onchain-apps/contracts/script/SignatureMintERC721.s.sol
```diff
@@ -0,0 +1,19 @@
...
+import {Script, console2} from "forge-std/Script.sol";
+import "../src/CustomERC1155.sol";
+import "../src/SignatureMintERC721.sol";
+
+contract SignatureMintERC721Script is Script {
+    function setUp() public {}
+
+    function run() public {
+        uint256 privateKey = vm.envUint("PRIVATE_KEY");
+        address signer = vm.envAddress("SIGNATURE_MINT_SIGNER");
```
Sneh1999, line 13: Wonder if `vm.envAddress` converts the address into the checksum address or do we need to it that? 

One user could potentially provide us an address in all lowercase and we might have to convert it to a checksum address is vm.envAddress doesnt do that for us

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+        freeMintSigner = signer;
+    }
+
+    /**
+     * @dev Function to mint for free using a wallet signature
+     * @param  account that will receive the NFT
+     * @param  signature that authorizes this transaction to mint for free.
+     */
+    function freeMint(address account, bytes memory signature) external {
+        require(this.verifySignature(account, signature) == true, "Unable to verify Signature");
```
Sneh1999, line 48: I believe this could just be `require(this.verifySignature(account, signature), "Unable to verify Signature");`

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+    constructor(address signer) ERC721(metadata.name, metadata.symbol) {
+        freeMintSigner = signer;
+    }
+
+    /**
+     * @dev Function to mint for free using a wallet signature
+     * @param  account that will receive the NFT
+     * @param  signature that authorizes this transaction to mint for free.
+     */
+    function freeMint(address account, bytes memory signature) external {
```
Sneh1999, line 47: I wonder if this should be an owner only function? Because otherwise a user can do a signature and send an account and get a free mint

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+        _mint(account, ++currentId);
+        emit Mint(account, currentId, 0);
+    }
+
+    /**
+     * @dev Function to mint for a fixed cost defined in the paidMintCost parameter
+     * @param  account that will receive the NFT
+     */
+    function mint(address account) external payable {
+        if (msg.value != paidMintCost) {
```
Sneh1999, line 64: What if they send a little extra? I think they should still be allowed to mint

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
wilsoncusack, line 48: not really idiomatic to use `this` unless there's some name space conflict?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * @dev Function to mint for free using a wallet signature
+     * @param  account that will receive the NFT
+     * @param  signature that authorizes this transaction to mint for free.
+     */
+    function freeMint(address account, bytes memory signature) external {
+        require(this.verifySignature(account, signature) == true, "Unable to verify Signature");
+
+        // Check if the free mint has already been used (1 per wallet)
```
wilsoncusack, line 50: might want to make this configurable?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
wilsoncusack: don't need `== true`

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+import "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
+import {NFTUtilities} from "./NFTUtilities.sol";
+
+// Errors
+error InsufficientFunds();
+
+/**
+ * @title SignatureMintERC721
+ * @dev contract that allows a user to mint a ERC721 for free with a cryptographically signed message.
+ *      This is useful for mints where you want to allow users to mint for free based on a signature from
```
wilsoncusack, line 17: So here the user is sort of consenting to a free mint? not sure I totally understand the need for a signature? Is someone else paying the gas?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+ *      your backend API.  This is an alternative approach to a merkel tree which is fully on-chain.
+ */
+contract SignatureMintERC721 is ERC721 {
+    using ECDSA for bytes32;
+    using MessageHashUtils for bytes32;
+
+    // Contract Params
+    uint256 public paidMintCost = 0.0001 ether;
```
wilsoncusack, line 25: probably make a constructor arg

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+    address public freeMintSigner; // Wallet Address for the authorized wallet to distribute free mints.
+    uint256 public currentId = 0; // Current NFT id
```
wilsoncusack, line 27: nit: defaults to 0;
```suggestion
    uint256 public currentId;
```

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
wilsoncusack: Oh wait the signature is from some admin?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+        return string(abi.encodePacked("Free Mint: ", Strings.toHexString(uint256(uint160(account)), 20)));
+    }
+
+    /**
+     * @dev Verifies a message was signed by the freeMintSigner wallet.
+     * @param  account that will receive the NFT
+     * @param  signature authorizing the action
+     */
+    function verifySignature(address account, bytes memory signature) public view returns (bool) {
+        bytes32 ethSignedMessageHash = this.getBytesToSign(account);
```
wilsoncusack, line 137: ethSigned message usually refers to something ERC 191 compliant, which this is not?

wilsoncusack: but would probably be better if it followed the ERC 191 or ERC 712

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+            NFTUtilities.getEncodedStringPair("symbol", metadata.symbol, false),
+            NFTUtilities.getEncodedStringPair("image", metadata.image, false),
+            NFTUtilities.getEncodedStringPair("animation_url", metadata.animation_url, true),
+            "}"
+        );
+        return string(jsonMetadata);
+    }
+
+    function tokenURI(uint256 tokenId) public view override(ERC721) returns (string memory) {
+        require(tokenId <= currentId, "Token doesnt exist");
```
wilsoncusack, line 88: Weird condition where this check passes for tokenId 0 on creation. Maybe should be nextTokenID and strictly less than?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * @dev Function to mint for a fixed cost defined in the paidMintCost parameter
+     * @param  account that will receive the NFT
+     */
+    function mint(address account) external payable {
+        if (msg.value != paidMintCost) {
+            revert InsufficientFunds();
+        }
+        _mint(account, ++currentId);
```
wilsoncusack, line 67: If you do `nextTokenId`
```suggestion
        _mint(account, currentId++);
```

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+    function freeMint(address account, bytes memory signature) external {
+        require(this.verifySignature(account, signature) == true, "Unable to verify Signature");
+
+        // Check if the free mint has already been used (1 per wallet)
+        require(usedFreeMints[account] == 0, "Free mint already used");
+        usedFreeMints[account] = 1;
+
+        // Mint the NFT
+        _mint(account, ++currentId);
+        emit Mint(account, currentId, 0);
```
wilsoncusack, line 56: Normally transfers from the 0 address are indexed as mints and there is no special mint event. Do you need it?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
```diff
@@ -0,0 +1,143 @@
...
+ *      This is useful for mints where you want to allow users to mint for free based on a signature from
+ *      your backend API.  This is an alternative approach to a merkel tree which is fully on-chain.
+ */
+contract SignatureMintERC721 is ERC721 {
+    using ECDSA for bytes32;
+    using MessageHashUtils for bytes32;
+
+    // Contract Params
+    uint256 public paidMintCost = 0.0001 ether;
+    address public freeMintSigner; // Wallet Address for the authorized wallet to distribute free mints.
```
wilsoncusack, line 26: could be immutable.

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: yes.. basically this is showcasing how you can use a backend wallet to gate specific contract functionality .. pretty good read on TW : https://portal.thirdweb.com/glossary/signature-based-minting

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: For context, this project is mostly to educate users so I am trying to not over-complicate the contract.  Most of the collection metadata is hardcoded.

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: freeMintSigner is passed via constructor.. do you think we should hardcode in the contract itself?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: This function requires a signed message vended by the _freeMintSigner_ address.  Thus, a user could not get a signature without interacting with the backend (which has the _freeMintSigner_ private key).

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: agreed

robpolak: agreed

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: again, this is meant to be an example users can configure for themself.  I didn't want to make the constructor super busy for this example.  @Zizzamia  any opinions here if we want to make the contracts constructor super configurable?

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: I thought it would be nice to showcase how a custom event would work.. Also, I wanted to be able to show the mint event + price easily on the UI.  I know you can get that with transaction logs, just requires a bit more work.

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: IDK, I am mixed here, there is no reason for a user to send extra.  We could write logic to refund, however, I think that just adds additional exploitable surface for little gain:

` //refund extra payment
        uint256 refund = msg.value - cost;
        (bool sent, ) = msg.sender.call{value: refund}("");
        if (!sent) {
            revert UnableToRefund();
        }`

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: Is there something dangerous doing it this way?  I assume you mean we should be using a counter : 

e.g. https://github.com/coinbase/nft-dapp-starter-kit/blob/1510766f576e9e03f6a9bdc4c9237e05a8736319/smart-contracts/contracts/nft.sol#L180
```
    // ============ SUPPORTING FUNCTIONS ============
    function nextTokenId() private returns (uint256) {
        tokenCounter.increment();
        return tokenCounter.current();
    }

```

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: This uses the same naming that OpenZepplin uses : https://docs.openzeppelin.com/contracts/2.x/utilities... AFAIK that is following ERC191 as we are using MessageHashUtils.toEthSignedMessageHash(messageEncodedPacked); in  getBytesToSign

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
Sneh1999: okay makes sense!

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: for example, mint price, should that be configurable via constructor or just hardcoded in the contract.  Back to our earlier point, we aren't trying to compete with OpenZepplin here so I don't think we need to fully future-proof our contracts.

Zizzamia: Right right, and I think it goes back to also the point on "how long does it take to understand and update a contract". 

Mint price in this case is 0, would be the case this contract will be used on a case is not 0? And if that's the case, a configuration will make an easier upgrade path vs hardcoded.

My guess at least for price, have that configurable seems reasonable.

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
wilsoncusack: oops sorry missed that, is right!

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
wilsoncusack: was just re the issue I mentioned with tokenId 0

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
wilsoncusack: immutable args can be set in the constructor!

apps/build-onchain-apps/contracts/script/SignatureMintERC721.s.sol
robpolak: this converts into an address object under the covers, so we should be good.  Tested locally and works as expected:
image

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: hrm.. I can't see to get that to work:

image

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: Okay taking in all the points here, I think we should keep it simple for now.  It will be trivial for a developer to make these values configurable.  I think the readability is more important at this phase.  This is an easy one-way door if we want to revisit at a later point.

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: fixed in next pr

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: fixed

apps/build-onchain-apps/contracts/src/NFTUtilities.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+library NFTUtilities {
```
Zizzamia, line 4: @robpolak @Sneh1999 open question, can this also be part of OnchainKit?

As OnchainKit will have three types of utilities:
- CSS
- TS
- Solidity

apps/build-onchain-apps/contracts/src/SignatureMintERC721.sol
robpolak: fixed with immutable

apps/build-onchain-apps/contracts/src/NFTUtilities.sol
robpolak: Not sure this is generic enough to qualify.  Again, I don't want to be authoring solidity components they get widely imported.  This should be referenced as a guide and not a framework to import.

Zizzamia: Like
```
onchainkit/contracts/utils/
```

robpolak: Given, I don't think we want to take on that scope of work & we don't have the resources to audit to the depth I think we should.

Zizzamia: Maybe I need to pull this folder https://github.com/coinbase/build-onchain-apps/tree/main/apps/build-onchain-apps/src/onchain a bit up, as we can reuse it for other things as well. 🤔

Sneh1999: yea, I also dont think we should

Zizzamia (approved):
LFG ✨

PR #232, opened 2024-01-09
Convert to next 13 app folder

wespickett:
**What changed? Why?**
Moved from pages to app folder

**Notes to reviewers**
There's some notable changes:
- route is still folder paths, and next looks for a page.tsx file to generate a route
- Need to put "use client" at the top of the client component. Pages import a client component
- How metadata works has changed

**How has it been tested?**


apps/build-onchain-apps/app/buy-me-coffee/page.tsx
```diff
@@ -0,0 +1,13 @@
+import { generateMetadata } from '../../src/utils/generateMetadata';
+import BuyMeCoffeePage from '.';
```
Zizzamia, line 2: Page from index? 🤔 is it how the cool kids to this these days?

apps/build-onchain-apps/app/buy-me-coffee/page.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const metadata = generateMetadata({
```
Zizzamia, line 4: soooo, medatada just gets absorbed automatically? would you mind link a doc on this.

apps/build-onchain-apps/app/layout.tsx
```diff
@@ -0,0 +1,39 @@
+import '../src/global.css';
+
+import { Roboto_Mono, Inter } from 'next/font/google';
+import OnchainProviders from '../src/providers/OnchainProviders';
+import { initAnalytics } from '../src/utils/analytics';
+import type { Metadata } from 'next';
+
+const roboto = Roboto_Mono({
```
Zizzamia, line 8: This kind of hurts my eyes.... so next has some sort of cool library that makes some of this working like magic..... mmmmmmm

apps/build-onchain-apps/package.json
```diff
@@ -28,6 +28,7 @@
     "next": "13.5.6",
     "next-pwa": "5.6.0",
     "perfume.js": "9.2.0",
+    "pino-pretty": "^10.3.1",
```
Zizzamia, line 31: Hey Pino what's up.
What do we use this for?

apps/build-onchain-apps/app/buy-me-coffee/page.tsx
wespickett: Yep

https://nextjs.org/docs/app/building-your-application/optimizing/metadata

I'd say they made it worse since now it's pretty hard to get router info to include in the metadata, which is why we have to pass the pathname explicitly now

apps/build-onchain-apps/app/buy-me-coffee/page.tsx
wespickett: Ya so the page is like the server component, and then you import the client component into it
https://nextjs.org/docs/pages/building-your-application/upgrading/app-router-migration#step-4-migrating-pages

apps/build-onchain-apps/app/layout.tsx
wespickett: no it doesn't work like magic, which I also assumed. These fonts are actually broken now, so still we need to update something so they get picked up everywhere again...

apps/build-onchain-apps/package.json
wespickett: It's a transitive dep, but the build was failing since it was being required but wasn't in the node_modules. The package that requires it had a comment that we have to include it separately

Zizzamia (approved):
Big big changes, nice!

PR #233, opened 2024-01-10
fix: fonts with no delay

wespickett:
**What changed? Why?**
fixed fonts: removed hack and use next/font/google

**Notes to reviewers**

**How has it been tested?**


Zizzamia (approved):
Brilliant!!!

PR #234, opened 2024-01-10
chore: Move to git clone instead of tar for setup

Sneh1999:
**What changed? Why?**
Moving to git clone instead of tar download to support git submodules and foundry

**Notes to reviewers**

**How has it been tested?**


package.json
```diff
@@ -7,10 +7,8 @@
     "chalk": "^4.1.2",
     "commander": "^11.1.0",
     "figlet": "^1.7.0",
-    "got": "^11.8.5",
     "inquirer": "^8.2.5",
     "rimraf": "^5.0.5",
-    "tar": "4.4.18",
```
Zizzamia, line 13: Bye bye tar

Zizzamia (approved):
Niiice

PR #235, opened 2024-01-10
feat: footer component

renanmav:
**What changed? Why?**

Implement footer component based on Figma design.

**Notes to reviewers**
Screenshot 2024-01-10 at 17 01 55

**How has it been tested?**
Locally.


apps/build-onchain-apps/app/home/index.tsx
```diff
@@ -15,6 +16,7 @@ export default function HomePage() {
         <WhyUseIt />
         <Guide />
       </main>
+      <Footer />
```
Zizzamia, line 19: Can we make this part of the Layout? And have the footer component everywhere.

apps/build-onchain-apps/src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,75 @@
+import { GitHubLogoIcon } from '@radix-ui/react-icons';
+import NextLink from 'next/link';
+import { NavbarLink } from '../header/Navbar';
+
+export default function Footer() {
+  return (
+    <footer className="h-[22.1875rem] bg-[#141519] py-12">
```
Zizzamia, line 7: Two things:
1. let's try to have rem that has not extra digits, like the simple the better
2. bg, we should create in the `tailwind.config.ts` file shared bg color, and add this one there. So it's more clean to use it.

apps/build-onchain-apps/app/home/index.tsx
renanmav: I'm having a weird createContext function not defined when I try to render the footer component within layout.tsx. I believe this is caused by Radix. cc @wespickett 

image

I'm just going to add the component to all pages for now...

renanmav: After syncing with @wespickett, we made it work within layout.tsx by adding `'use client'` at the top of the footer component code. Check 3a93743.

apps/build-onchain-apps/src/components/footer/Footer.tsx
renanmav: @Zizzamia done at d0d419b

Zizzamia (approved):
Nice

PR #240, opened 2024-01-12
Fix: Fixes signature mint errors

robpolak:
**What changed? Why?**
1. Fixed the singnature mint page being broken due to a bad contract address configuration.
2. Added margin to the top of the signature mint page
3. Added/Enhanced text on the signature mint page.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/apps/build-onchain-apps/src/components/mint/SignatureMint.tsx
+++ b/apps/build-onchain-apps/src/components/mint/SignatureMint.tsx
@@ -36,6 +36,8 @@ const codeBlock3 = `# Edit the .env file and add the following key:
 SIGNATURE_MINT_SIGNER="WalletAddressFromAbove"
 `;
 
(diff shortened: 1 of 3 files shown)


cnasc (approved):
LGTM

apps/build-onchain-apps/src/components/mint/SignatureMint.tsx
```diff
@@ -36,6 +36,8 @@ const codeBlock3 = `# Edit the .env file and add the following key:
 SIGNATURE_MINT_SIGNER="WalletAddressFromAbove"
 `;
 
+const codeBlock4 = `curl 'http://localhost:3000/api/mint/signature/free?chainId=84532&wallet=0xWALLET_ADDRESS'`;
```
Zizzamia, line 39: Let's use ENV variable here instead of localhost, so that it works when deployed on vercel as well.

Zizzamia: Actually nver mind.

PR #251, opened 2024-01-16
fix: remove theme null

wespickett:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


Zizzamia (approved):
Great catch!!!

PR #268, opened 2024-01-23
Refactor: Move APIs from Page Router to App Router

robpolak:
**What changed? Why?**
1. Refactored the API backend from the pages router to app router.
2. I did notice some of these errors (but the build runs and passes)

`Generating static pages (0/10)  [    ]Error fetching chains: eS [Error]: Dynamic server usage: Page couldn't be rendered statically because it used `nextUrl.searchParams`. See more info here: https://nextjs.org/docs/messages/dynamic-server-error`

Fixed this with : [REF](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic)
```export const dynamic = 'force-dynamic';```

**Notes to reviewers**
1. I was unable to get to unit tests working.  Ran into issues mocking next/server, going to dig in another time when I have some more capacity.
3. I figure getting the pattern is more critical than unit tests, so getting this merged up so we can move forward.

**How has it been tested?**
1. Tested all of the API routes + Signature Mint pages.

### CODE REVIEW

File: template/app/api/chains/supported/route.ts [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+  try {
+    const chains = getChainsForEnvironment();
+    return NextResponse.json(chains, { status: 200 });
+  } catch (error) {
+    console.error('Error fetching chains:', error);
+    return NextResponse.json({}, { status: 500, statusText: 'Internal Server Error' });
+  }
+}
+
+export const dynamic = 'force-dynamic';
```
Sneh1999, line 20: Wondering why do we want this to be dynamic? List of supported chains wouldnt change a lot IMO

template/src/pages/signature-mint/ContractDemo.tsx
```diff
@@ -138,8 +138,8 @@ export default function SignatureMintDemo() {
             {sigFailure && (
               <p className="mb-4 rounded-lg bg-yellow-100 p-4 text-sm text-yellow-700" role="alert">
                 <span className="font-medium">Warning!</span> There was a problem with the
-                signature. Please follow the steps below and ensure SIGNATURE_MINT_SIGNER is set in
-                the environment file.
+                signature. Please follow the steps below and ensure SIGNATURE_MINT_PRIVATE_KEY is
```
Sneh1999, line 141: Shoul update the `.env.example` file as well and also the `contracts/README.md` file :)

Sneh1999 (commented):
LGTM! just a few comments

template/app/api/chains/supported/route.ts
Sneh1999, line 20: @rob.polak I was looking through the code, I couldnt find where we were using nextURL.searchParams for /api/chains/supported. Maybe we could remove this line for this route? 
I am suggesting this because I think the supported chains would not change, so we might get some caching benefits. What are your thoughts?

template/src/pages/signature-mint/ContractDemo.tsx
robpolak: This is only valid for the FE.. We need the private key to sign on the FE and the address to validate within the contract.

robpolak: So in contract we need SIGNATURE_MINT_SIGNER

In the FE we need SIGNATURE_MINT_PRIVATE_KEY

Sneh1999: Arent all variables/env variables on frontend public? Wouldnt this be a security flaw then

Sneh1999: Arent all variables/env variables on frontend public? Wouldnt this be a security flaw then

template/app/api/chains/supported/route.ts
wespickett: @robpolak @Sneh1999 I think the 'force-dynamic' can also relate to caching, so if the getChainsForEnvironment is considered dynamic (will return something new each time) then we would want to keep this dynamic

Sneh1999: I think we should have caching on this route as the list of supported chains currently doesnt change a lot, open to thoughts!

Zizzamia: Side note, we should document in some way how force-dynamic works, or at least have link for public docs.

Zizzamia:
Dope

PR #271, opened 2024-01-24
Fix foundry-related configuration

gabrielstoica:
**What changed? Why?**

1. Fix typo in `foundry.toml` file.
2. Add `url` param to Etherscan config as per [Foundry documentation](https://book.getfoundry.sh/reference/config/etherscan). Upon running `forge script script/BuyMeACoffee.s.sol --rpc-url base_sepolia` we'll get the following error: "At least one of 'url' or 'chain' must be present for Etherscan config with unknown alias 'base_sepolia'", hence adding the `url` of the explorer will fix it.
3. Use the `base_sepolia` alias directly in the TOML file instead of referencing it to avoid any errors.

### DIFF
--- a/template/contracts/.env.example
+++ b/template/contracts/.env.example
@@ -3,3 +3,4 @@ RPC_URL="https://sepolia.base.org"
 NETWORK="base_sepolia"
 SIGNATURE_MINT_SIGNER=
 BLOCK_EXPLORER_API_KEY="GET_API_KEY_FROM_THE_EXPLORER"
+BLOCK_EXPLORER="https://sepolia.basescan.org/"
--- a/template/contracts/foundry.toml
+++ b/template/contracts/foundry.toml
@@ -6,9 +6,9 @@ fs_permissions = [{ access = "read", path = "./"}]
 solc="0.8.23"
 
 [rpc_endpoints]
-"${NETWORK}"="${RPC_URL}"
+base_sepolia="${RPC_URL}"
 
 [etherscan]
-"${NETWORK}"={key="${BLOCK_EXPLORER_API_KEY}",url="${BLOCK_EXPLORER}"}
+base_sepolia={key="${BLOCK_EXPLORER_API_KEY}",url="${BLOCK_EXPLORER}"}
 
 # See more config options https://github.com/foundry-rs/foundry/tree/master/config
--- a/template/contracts/foundry.toml
+++ b/template/contracts/foundry.toml
@@ -6,7 +6,7 @@ fs_permissions = [{ access = "read", path = "./"}]
 solc="0.8.23"
 
 [rpc_endpoints]
-"${NETWORK}"="${RPC_URl}"
+"${NETWORK}"="${RPC_URL}"
 
 [etherscan]
 "${NETWORK}"={key="${BLOCK_EXPLORER_API_KEY}"}
--- a/template/contracts/foundry.toml
+++ b/template/contracts/foundry.toml
@@ -9,6 +9,6 @@ solc="0.8.23"
 "${NETWORK}"="${RPC_URL}"
 
 [etherscan]
-"${NETWORK}"={key="${BLOCK_EXPLORER_API_KEY}"}
+"${NETWORK}"={key="${BLOCK_EXPLORER_API_KEY}",url="${BLOCK_EXPLORER}"}
 
 # See more config options https://github.com/foundry-rs/foundry/tree/master/config


template/contracts/foundry.toml
```diff
@@ -6,9 +6,9 @@ fs_permissions = [{ access = "read", path = "./"}]
 solc="0.8.23"
 
 [rpc_endpoints]
-"${NETWORK}"="${RPC_URl}"
+base_sepolia="${RPC_URL}"
 
 [etherscan]
-"${NETWORK}"={key="${BLOCK_EXPLORER_API_KEY}"}
+base_sepolia={key="${BLOCK_EXPLORER_API_KEY}",url="${BLOCK_EXPLORER}"}
```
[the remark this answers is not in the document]
Sneh1999: It allows for any network and you can specify the network in `.env`

template/contracts/.env.example
```diff
@@ -3,3 +3,4 @@ RPC_URL="https://sepolia.base.org"
 NETWORK="base_sepolia"
 SIGNATURE_MINT_SIGNER=
 BLOCK_EXPLORER_API_KEY="GET_API_KEY_FROM_THE_EXPLORER"
+BLOCK_EXPLORER="https://sepolia.basescan.org/"
```
Sneh1999, line 6: I believe you dont need to specify the BLOCK_EXPLORER, foundry is able to figure it out with your API KEY

template/contracts/foundry.toml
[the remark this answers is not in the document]
gabrielstoica: Got it, but it's not a good practice to restrict the user to using only one chain/configuration at a time. If I want to deploy on multiple chains, I'm forced to switch the network in `.env` every time. The user should be allowed to add as many networks as they want, as per the Foundry book documentation: https://book.getfoundry.sh/reference/config/etherscan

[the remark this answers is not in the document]
gabrielstoica: You'll also get a `bad fork_url provider: base_sepolia` when running for example `forge script script/BuyMeACoffee.s.sol --rpc-url $NETWORK`

template/contracts/.env.example
gabrielstoica: Since`base_sepolia` is an unknown alias for Foundry, when you're going to run any script, it will require to add at least one of `url` or `chain` to the Etherscan config. Again, check the Foundry book documentation: https://book.getfoundry.sh/reference/config/etherscan#etherscan

gabrielstoica: This it the error you're going to get: `"At least one of 'url' or 'chain' must be present for Etherscan config with unknown alias 'base_sepolia'"`

Zizzamia: Thank you for sharing this @gabrielstoica.
Just out of curiosity, btw when in particular this error is happening?

gabrielstoica: Don't worry @Zizzamia!

Considering the following configuration: 
```
[rpc_endpoints]
base_sepolia="${RPC_URL}"

[etherscan]
base_sepolia={key="${BLOCK_EXPLORER_API_KEY}"}
```

You'll get that error when running any script such as: `forge script script/BuyMeACoffee.s.sol --rpc-url base_sepolia`

template/contracts/foundry.toml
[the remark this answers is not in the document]
Zizzamia: Yeah makes sense.

@gabrielstoica do you have a good example of `foundry.toml` you liked in the past that we can take a look.

Also if you are thinking an end-to-end experience from coding, to deploy and so on. How you would like to approach this?

And again, thank you for taking the time to share your thoughts.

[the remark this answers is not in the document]
gabrielstoica: Sure, you can have a look at this `foundry.toml` - it has a lot of useful configurations: https://github.com/PaulRBerg/foundry-template/blob/main/foundry.toml.

As stated before, I'd suggest to allow the user to use as many networks as they want by removing the `$NETWORK` restriction from `.env`. It's also worth looking at the `.env.example` file within that repository and following that pattern.

On another note, everything else looks good to me. @Zizzamia

template/contracts/.env.example
Sneh1999: @gabrielstoica  Oh I see, I know why you are getting this error. You need to update your foundry version, they have made base_sepolia an officia alias now. 

You will not get any of the errors you have been happening and everything will work :)

Sneh1999: `base_sepolia` was recently introduced and a reason foundry's older version doesnt suppport it

template/contracts/foundry.toml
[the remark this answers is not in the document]
Sneh1999: okay makes sense! @gabrielstoica `bad fork_url provider: base_sepolia` is also happening because you are prob on an older version of foundry, let me know if that is not the case.

When it comes to allowing multiple chains at a time, yes I agree. That is something, we can def try doing. 

Maybe we can restructure your PR for that and support `base mainnet` and `base sepolia` at the same time.

gabrielstoica:
Hi @Zizzamia, @Sneh1999!
Base Mainnet, Base Goerli and Base Sepolia now have aliases that can be used directly in the `foundry.toml` file. This also eliminates the need to append `--etherscan-api-key ${BLOCK_EXPLORER_API_KEY}` to each `forge script` call.  

FYI: https://github.com/alloy-rs/chains/pull/32.

Sneh1999 (approved):
LGTM! @gabrielstoica great work and thanks again for the contribution.

gabrielstoica:
Thanks @Sneh1999, happy to contribute!

Zizzamia:
Brilliant!
Thank you @gabrielstoica @mattsse @DaniPopes and @Evalir

PR #272, opened 2024-01-24
Alvaroraminelli/ens hooks caching

alvaroraminelli:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


template/onchainKit/hooks/useEnsData.ts
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from 'react';
+import { GetEnsAvatarReturnType, GetEnsNameReturnType } from 'viem/ens';
+import { StorageInterface } from '../types';
+// The fetcher function type, which can receive multiple arguments
+type FetcherFunction<T extends unknown[]> = (...args: T) => Promise<GetEnsNameReturnType | GetEnsAvatarReturnType | undefined>;
+
+// useEnsData: A generic hook to fetch and store data using a specified storage service.
```
Zizzamia, line 7: neat, comments for function should be
```
/**
 *
 */
```

template/onchainKit/hooks/useEnsData.ts
```diff
@@ -0,0 +1,44 @@
...
+// It fetches data based on the given dependencies and stores it using the provided storage service.
+export function useEnsData<T extends unknown[]>(
+  fetcher: FetcherFunction<T>, 
+  deps: T, 
+  storageService: StorageInterface
+) {
+  const [data, setData] = useState<string | null | undefined>(undefined);
```
Zizzamia, line 14: let's name this type

template/onchainKit/hooks/useEnsData.ts
```diff
@@ -0,0 +1,44 @@
...
+import { GetEnsAvatarReturnType, GetEnsNameReturnType } from 'viem/ens';
+import { StorageInterface } from '../types';
+// The fetcher function type, which can receive multiple arguments
+type FetcherFunction<T extends unknown[]> = (...args: T) => Promise<GetEnsNameReturnType | GetEnsAvatarReturnType | undefined>;
+
+// useEnsData: A generic hook to fetch and store data using a specified storage service.
+// It fetches data based on the given dependencies and stores it using the provided storage service.
+export function useEnsData<T extends unknown[]>(
+  fetcher: FetcherFunction<T>, 
+  deps: T,
```
Zizzamia, line 11: what's deps?
just in general let's have very descriptive variables

template/onchainKit/hooks/useEnsData.ts
```diff
@@ -0,0 +1,44 @@
...
+  deps: T, 
+  storageService: StorageInterface
+) {
+  const [data, setData] = useState<string | null | undefined>(undefined);
+
+  useEffect(() => {
+    let isSubscribed = true;
+
+    const checkAndFetchData = async () => {
+      try {
```
Zizzamia, line 20: this is quite a big try/catch, is usually a best practice to keep try catch small, as they tend to hold memory inside the browser.

Zizzamia: like should the try catch just leave inside the storageService

template/onchainKit/hooks/useEnsData.ts
```diff
@@ -0,0 +1,44 @@
...
+        const id = deps.join('-');
+        const cachedData = await storageService.getData(id);
+
+        if (cachedData && isSubscribed) {
+          setData(cachedData);
+        } else {
+          const fetchedData = await fetcher(...deps);
```
Zizzamia, line 27: I will rather we are more explicit on what are `deps`, in general when writing a library we should almost never use `...` as that will hide complexity and makes the code harder to mantain and understand.

There are case it's fair to use `...` but it should be more like a necessicy that a nice to have.

template/onchainKit/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,23 @@
+import { StorageInterface } from "../types";
+
+// InMemoryStorageService: A service to store data in JavaScript memory.
+// It's a simple key-value storage using a Map. The data is stored as long as the page is not reloaded.
+export class InMemoryStorageService implements StorageInterface {
+    private storage: Map<string, string | null | undefined>;
```
Zizzamia, line 6: let's name those types

template/onchainKit/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    constructor() {
+        this.storage = new Map<string, string | null | undefined>();
```
Zizzamia, line 9: run format, you have 4 spaces instead of 2

template/onchainKit/hooks/useEnsName.ts
```diff
@@ -1,34 +1,27 @@
...
+  } catch (err) {
+    return undefined;
+  }
+};
 
+/**
+ * Fetches the ENS name for a given address.
+ */
+export const useEnsName = (address: Address) => {
+  const ensActionKey =  address ?? '';
```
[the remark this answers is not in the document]
alvaroraminelli: ideally we should not even call the action in such case.

template/onchainKit/hooks/useOnchainActionWithCache.ts
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from 'react';
+
+import { StorageInterface, ActionFunction, ActionResponse, ActionKey } from '../types';
+
+/**
+ * A generic hook to fetch and store data using a specified storage service.
+ * It fetches data based on the given dependencies and stores it using the provided storage service.
+ */
+
```
Zizzamia, line 9: no need of this extra space

template/onchainKit/types.ts
```diff
@@ -1,6 +1,20 @@
+import { GetEnsAvatarReturnType, GetEnsNameReturnType } from 'viem/ens';
+
 export type WagmiFetchBalanceResult = {
```
Zizzamia, line 3: Are we still using this? Probably not

template/onchainKit/hooks/useEnsName.ts
```diff
@@ -1,34 +1,27 @@
...
+  const ensName = useOnchainActionWithCache(ensNameAction(address), ensActionKey, inMemoryStorageService);
```
Zizzamia, line 25: What our thinking around `ensActionKey`? 

Like should this be a key like `ens-name-${address}`?

template/onchainKit/hooks/useEnsName.ts
Zizzamia, line 24: Why we allow an empty state? isn't the type not allowing this?

template/onchainKit/store/inMemoryStorageService.ts
Zizzamia: should the type just be `string | string` ?

template/onchainKit/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,28 @@
...
+            return null;            
+        }
+    }
+
+    async setData(key: string, value: string | null | undefined): Promise<void> {
+        if (value === null) {
+            this.storage.delete(key);
+        } else {
+            this.storage.set(key, value);
+        }
```
Zizzamia, line 26: We miss a try catch for setting

template/onchainKit/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,28 @@
...
+        try {
+            return this.storage.has(key) ? this.storage.get(key) : null;
+        } catch (error) {
+            console.error('error reading data from storage', error);
+            return null;            
+        }
+    }
+
+    async setData(key: string, value: string | null | undefined): Promise<void> {
+        if (value === null) {
```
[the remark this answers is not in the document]
alvaroraminelli: the goal here is remove records with value null from the store, so we can save some space in the storage.

template/onchainKit/types.ts
alvaroraminelli: yes we are in balance utils

template/onchainKit/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,32 @@
+import { StorageInterface } from '../types';
+
+// InMemoryStorageService: A service to store data in JavaScript memory.
```
Zizzamia, line 3: neat. missing multi-line comments

template/onchainKit/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,32 @@
...
+// It's a simple key-value storage using a Map. The data is stored as long as the page is not reloaded.
+export class InMemoryStorageService implements StorageInterface {
```
Zizzamia, line 5: Not for this PR, BUT, I even wonder if we need this class, and we should just make this a functional-oriented code.

Mostly because Functional-orinted is more memory efficient. And also there is less risk to have this object with multiple instances.

Zizzamia (approved):
Well done!

PR #279, opened 2024-01-25
refactor: Header styles into tailwind

n3ps:
**What changed? Why?**
Migrate Header styles to Tailwind, instead of a mix of css modules and classes

**Notes to reviewers**
Also fixes this header layout issue:

BEFORE

Screenshot 2024-01-25 at 4 26 51 PM

AFTER
Screenshot 2024-01-25 at 4 29 22 PM

**How has it been tested?**
Locally


template/src/components/header/Header.tsx
```diff
@@ -1,6 +1,5 @@
 import { useState, useEffect } from 'react';
 import { clsx } from 'clsx';
-import styles from './Header.module.css';
```
Zizzamia, line 3: Bye bye `Header.module.css`

Zizzamia (approved):
BRILLIANT

PR #283, opened 2024-01-26
Utility to see if a user can afford amount + utilize in FormBuyCoffee

GioLogist:
**What changed? Why?**
- New utility for `useAddressCanAfford`
- New utility for `useLoggedInUserCanAfford`
- Unit tests for both
- Utilize above utilities to adjust UX if a user cannot afford to buy coffee

**Notes to reviewers**
## Before
If a user  would be allowed to initiate a transaction, which would lead to a failure of transaction, without notifying the user

## After
The UX adjusts accordingly
Screenshot 2024-01-27 at 9 07 29 AM

**How has it been tested?**

- Visit `/buy-me-coffee` with an account that has > 0.5ETH on Base Sepolia. User should be able to buy coffee
- Visit `/buy-me-coffee` with an account that has < 0.5ETH on Base Sepolia. User should **not** be able to buy coffee

### DIFF
--- a/template/src/pageComponents/buy-me-coffee/FormBuyCoffee.tsx
+++ b/template/src/pageComponents/buy-me-coffee/FormBuyCoffee.tsx
@@ -10,7 +10,7 @@ type FormBuyCoffeeProps = {
   onComplete: () => void;
 };
 
-const BUY_COFFEE_AMOUNT_RAW = '0.5';
+const BUY_COFFEE_AMOUNT_RAW = '0.0001';
 const BUY_COFFEE_AMOUNT = parseEther(BUY_COFFEE_AMOUNT_RAW);
 
 function FormBuyCoffee({ onComplete }: FormBuyCoffeeProps) {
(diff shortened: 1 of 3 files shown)


template/src/pageComponents/buy-me-coffee/FormBuyCoffee.tsx
```diff
@@ -2,13 +2,23 @@ import React, { useState, useCallback, useMemo } from 'react';
...
+} from 'wagmi';
 import { useBuyMeACoffeeContract } from '../../hooks/contracts';
+import { useBalance } from 'wagmi';
+import { useLoggedInUserCanAfford } from '../../hooks/useUserCanAfford';
 
 type FormBuyCoffeeProps = {
   onComplete: () => void;
 };
 
+const BUY_COFFEE_AMOUNT_RAW = '0.5';
```
Zizzamia, line 19: That's an expensive coffee LOL.

I believe the coffee we now is `0.0001`, so `0.0002` should be enough.

GioLogist: LOL good catch. Just reverted to 0.0001. We'll revisit the price in a couple weeks based on value added 😂

Zizzamia (approved):
BRILLIANT

PR #311, opened 2024-02-01
chore: Detect out of gas

Sneh1999:
**What changed? Why?**
Add code to detect out of gas state

**Notes to reviewers**

**How has it been tested?**


arsood (approved):
Great! Thank you :)

PR #327, opened 2024-02-06
feat: Mint NFT Stepper

arsood:
**What changed? Why?**

- Added stepper functionality to Mint NFT flow
- Add callbacks for success and error on contract write
- Standardized Button component

**Screenshots**

_Start Mint NFT_

Screenshot 2024-02-05 at 11 12 03 PM

_Mint Processing_

Screenshot 2024-02-05 at 11 12 10 PM

_Out Of Gas_

Screenshot 2024-02-05 at 11 13 11 PM

_Mint Complete_

Screenshot 2024-02-05 at 11 13 40 PM

**How has it been tested?**

- Locally


template/src/pageComponents/mint/ContractDemo.tsx
```diff
@@ -1,16 +1,29 @@
...
 import { useCustom1155Contract } from '../../hooks/contracts';
 import NotConnected from './NotConnected';
+import MintCompleteStep from './steps/MintCompleteStep/MintCompleteStep';
+import MintProcessingStep from './steps/MintProcessingStep/MintProcessingStep';
+import OutOfGasStep from './steps/OutOfGasStep/OutOfGasStep';
+import StartMintStep from './steps/StartMintStep/StartMintStep';
 import SwitchNetwork from './SwitchNetwork';
 
-const EXPECTED_CHAIN = baseSepolia;
+export const EXPECTED_CHAIN = baseSepolia;
```
Zizzamia, line 15: I wonder if we should have some sort of config file for this.

Or even better question, should this information come from the Hook that was generated?

template/src/pageComponents/mint/steps/MintCompleteStep/MintCompleteStep.tsx
```diff
@@ -0,0 +1,41 @@
...
+        It will take ~ 5 minutes to show up in your wallet
+      </div>
+
+      <button
+        type="button"
+        className="block w-full rounded-full bg-white py-4 text-center text-sm text-black"
+        onClick={handleMintAnother}
+      >
+        Mint another NFT
+      </button>
```
Zizzamia, line 38: I feel like we should standardize this button style/logic into `components/Button.tsx` so we can keep the same logic style everywhere.

By doing that, if someone wants to than replace it with their own Design System, is relativelly much faster.

template/src/pageComponents/mint/steps/MintProcessingStep/MintProcessingStep.tsx
```diff
@@ -0,0 +1,33 @@
+import { SymbolIcon } from '@radix-ui/react-icons';
+import clsx from 'clsx';
+
+export default function MintProcessingStep() {
```
Zizzamia, line 4: I see we create this Issue for later https://github.com/coinbase/build-onchain-apps/issues/328, which is great.

I was thinking if we should also find a way to consolidate or abstract how states work in a contract. So it's easier for a developer re-build them.

template/src/pageComponents/mint/steps/OutOfGasStep/OutOfGasStep.tsx
```diff
@@ -0,0 +1,40 @@
+import { useCallback } from 'react';
+import clsx from 'clsx';
+import { MintSteps } from '../../ContractDemo';
+
```
Zizzamia, line 4: Open question do we want all this levels?
`steps/OutOfGasStep/OutOfGasStep.tsx`

or just have all steps file under `steps/OutOfGasStep.tsx`

arsood: That's a good point. Initially I added them under their own folders in case we needed other files there, but we can have this more top-level. I will make that change.

template/src/pageComponents/mint/ContractDemo.tsx
```diff
@@ -23,18 +36,21 @@ export default function MintContractDemo() {
...
-    functionName: 'mint',
-    args: address ? [address, BigInt(1), BigInt(1), address] : undefined,
-    enabled: onCorrectNetwork,
-  });
+  const mintContent = useMemo(() => {
+    if (mintStep === MintSteps.MINT_PROCESSING_STEP) {
+      return <MintProcessingStep />;
+    }
+
+    if (mintStep === MintSteps.OUT_OF_GAS_STEP) {
```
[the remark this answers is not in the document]
arsood: It's a good question, and something I was thinking a lot about. I decided against abstracting them even though there is similarity due to the fact that we may have complex logic in these steps at some point that are specific to the particular action (Mint, Buy Coffee, etc.). If we continue to see overlap it may be worth abstracting in a future PR.

template/src/pageComponents/mint/steps/MintCompleteStep/MintCompleteStep.tsx
arsood: Good idea. I'll do this.

template/src/pageComponents/mint/steps/OutOfGasStep/OutOfGasStep.tsx
Zizzamia: My point in general with BOAT, is to keep things as flat as possible, as than the developer that will eventually use it, can decide how to re-organizite after.

Zizzamia (approved):
LFG

template/yarn.lock
```diff
@@ -8,28 +8,28 @@ __metadata:
 "@aashutoshrathi/word-wrap@npm:^1.2.3":
   version: 1.2.6
   resolution: "@aashutoshrathi/word-wrap@npm:1.2.6"
-  checksum: 10c0/53c2b231a61a46792b39a0d43bc4f4f776bb4542aa57ee04930676802e5501282c2fc8aac14e4cd1f1120ff8b52616b6ff5ab539ad30aa2277d726444b71619f
+  checksum: 53c2b231a61a46792b39a0d43bc4f4f776bb4542aa57ee04930676802e5501282c2fc8aac14e4cd1f1120ff8b52616b6ff5ab539ad30aa2277d726444b71619f
```
eragon512, line 11: is this `yarn.lock` change intentional?

template/src/pageComponents/mint/steps/MintCompleteStep/MintCompleteStep.tsx
arsood: Done :)

template/src/pageComponents/mint/ContractDemo.tsx
arsood: Added as constant for now. We can refactor to hook in follow up PR.

PR #332, opened 2024-02-07
Component Cleanup and Button Component

arsood:
**What changed? Why?**

- Cleaned up components for Buy Me a Coffee and Mint experiences
- Flatten Buy Me a Coffee component tree
- Added Button component and implemented it across the two flows

**How has it been tested?**

- Locally


template/src/components/Button/Button.tsx
```diff
@@ -3,6 +3,7 @@ import clsx from 'clsx';
 
 type ButtonProps = {
   buttonContent: ReactNode | string;
+  isSubmit?: boolean;
```
n3ps, line 6: Suggestion passing down `type` will be more flexible

arsood: TypeScript throws errors with any use of type beyond static string and ternary expression. Ugh.

arsood: @n3ps I resolved this by using the type and disabling the Eslint rule

Zizzamia (approved):
Niiiice

PR #329, opened 2024-02-06
feat: add mobile navbar

eragon512:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Manual + Unit tests

|Before|After|
|---|---|
|Screenshot 2024-02-07 at 10 51 55 AM|Screenshot 2024-02-07 at 10 53 24 AM|
|Screenshot 2024-02-07 at 10 54 18 AM|Screenshot 2024-02-07 at 10 54 46 AMScreenshot 2024-02-07 at 10 55 32 AM|

PR #337, opened 2024-02-08
chore: cleanup header, Navbar

eragon512:
**What changed? Why?**
Follow-up on comments from PR #329 

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/template/src/components/header/AccountInfoPanel.tsx
+++ b/template/src/components/header/AccountInfoPanel.tsx
@@ -16,20 +16,14 @@ export function AccountInfoPanel() {
       <div className="my-4 inline-flex items-center justify-start gap-2">
         <OnchainAvatar address={address} className="h-10 w-10 rounded-full" />
         <div className="inline-flex flex-col items-start justify-center gap-1">
-          <div className="inline-flex items-center justify-start gap-1">
-            <div className="font-inter w-32 text-base font-medium text-white">{ensName}</div>
-          </div>
-          <div className="inline-flex items-center justify-start gap-8">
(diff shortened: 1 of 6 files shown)


template/src/components/header/AccountInfoPanel.tsx
```diff
@@ -16,20 +16,14 @@ export function AccountInfoPanel() {
...
-          </div>
+          <div className="font-inter w-32 text-base font-medium text-white">{ensName}</div>
+          <OnchainAddress
+            address={address}
+            className="font-inter w-32 text-sm font-medium text-zinc-400"
+          />
         </div>
       </div>
-      <div className="h-px self-stretch bg-zinc-400 bg-opacity-20" />
+      <hr className="h-px self-stretch bg-zinc-400 bg-opacity-20" />
```
arsood, line 26: Why the change from div to hr? Would prefer a div over an hr.

eragon512: In this case, this element is supposed to render a separator between the onchain address and the logout button

Having this be an hr element instead of a seems to fit better from a Semantic HTML POV, and the feedback from the earlier PR was to use more Semantic HTML

eragon512: Link to the feedback: https://github.com/coinbase/build-onchain-apps/pull/329#discussion_r1481025596

arsood:
@eragon512 let's run prettier to get the build passing and then we can approve and merge.

eragon512:
@arsood updated :+1:

arsood (approved):

PR #355, opened 2024-02-12
Fetch free 1155 mint gas estimation

Yuripetusko:
**What changed? Why?**
Fetch free 1155 mint gas estimation to display it to user instead of hardcoded value

**Notes to reviewers**

**How has it been tested?**
- `yarn test`
- `manual in-browser testing`

### DIFF
--- a/template/src/pageComponents/mint/ContractDemo.tsx
+++ b/template/src/pageComponents/mint/ContractDemo.tsx
@@ -1,5 +1,6 @@
 import { useMemo, useState } from 'react';
-import { useAccount } from 'wagmi';
+import { encodeFunctionData, formatEther } from 'viem';
+import { useAccount, useEstimateGas } from 'wagmi';
 import { useCollectionMetadata } from '../../../onchainKit';
 import NextImage from '../../components/NextImage/NextImage';
 import { EXPECTED_CHAIN } from '../../constants';
@@ -18,7 +19,7 @@ export enum MintSteps {
 export default function MintContractDemo() {
   const [mintStep, setMintStep] = useState<MintSteps | null>(null);
 
-  const { chain } = useAccount();
+  const { chain, address } = useAccount();
 
   const contract = useCustom1155Contract();
 
@@ -29,6 +30,23 @@ export default function MintContractDemo() {
     contract.status === 'ready' ? contract.address : undefined,
     contract.abi,
   );
+
+  // Estimate free minting 1 erc1155 NFT
+  const { data: txFeeEstimation, isLoading: isLoadingFeeEstimate } = useEstimateGas({
+    to: contract.status === 'ready' ? contract.address : undefined,
+    account: address,
+    chainId: chain?.id,
+    data: address
+      ? encodeFunctionData({
+          abi: contract.abi,
+          functionName: 'mint',
+          args: [address, BigInt(1), BigInt(1), address],
+        })
+      : undefined,
+    query: { enabled: onCorrectNetwork && !!address },
+  });
+
+  const mintTxFeeEstimation = txFeeEstimation ? formatEther(txFeeEstimation, 'gwei') : undefined;
 
   const mintContent = useMemo(() => {
     return (
@@ -48,13 +66,10 @@ export default function MintContractDemo() {
     return <SwitchNetwork />;
   }
 
-  if (isLoading) {
+  if (isLoading || isLoadingFeeEstimate) {
     // A future enhancement would be a nicer spinner here.
     return <span className="text-xl">loading...</span>;
   }
-
-  // TODO: Retrieve this dynamically
-  const ethAmount = 0.0001;
 
   return (
     <div className="my-10 gap-16 lg:my-20 lg:flex">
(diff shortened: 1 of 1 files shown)


template/src/pageComponents/mint/ContractDemo.tsx
```diff
@@ -68,7 +83,11 @@ export default function MintContractDemo() {
       <div className="flex-shrink-1 mt-10 w-full flex-grow-0 lg:mt-0">
         <h1 className="text-4xl font-bold">{collectionName}</h1>
 
-        <h2 className="my-5">{String(ethAmount)} ETH</h2>
+        {mintTxFeeEstimation && (
+          <h2 className="my-5">
+            Estimated tx fee: ~{mintTxFeeEstimation} {chain?.nativeCurrency.symbol ?? 'ETH'}
+          </h2>
+        )}
```
kyhyco, line 90: Haven't played much with `wagmi` but we could be more descriptive here. Something like:

 ```suggestion
        <h2 className="my-5">
            {mintTxFeeEstimation 
              ?  `Estimated tx fee: ~${mintTxFeeEstimation} ${chain?.nativeCurrency.symbol ?? 'ETH'}`
              :  "Unable to estimate" }
        </h2>
```

Since `isLoadingFeeEstimate` is `false`, `mintTxFeeEstimation` should have a returned value.

Comment (Yuripetusko) [reply template/src/pageComponents/mint/ContractDemo.tsx]: How about now?. I am not hiding the whole page now just because tx fee is loading, and added a simple loading spinner.

File: template/src/pageComponents/mint/ContractDemo.tsx [hunk 2]
```diff
@@ -48,14 +72,16 @@ export default function MintContractDemo() {
...
 
-  if (isLoading) {
-    // A future enhancement would be a nicer spinner here.
-    return <span className="text-xl">loading...</span>;
+  if (isLoadingCollectionMetadata || contract.status !== 'ready') {
+    return (
+      <div className="my-5 flex justify-center align-middle">
+        <span className="text-xl">
+          <SpinnerIcon className="h-20 w-20 animate-spin" />
+        </span>
```
Sneh1999, line 80: I think SpinnerIcon can be more of a utility and not specific to this file as it can be used in other places as well

template/src/pageComponents/mint/ContractDemo.tsx
```diff
@@ -68,7 +94,17 @@ export default function MintContractDemo() {
...
         <h1 className="text-4xl font-bold">{collectionName}</h1>
 
-        <h2 className="my-5">{String(ethAmount)} ETH</h2>
+        <h2 className="my-5">
+          Estimated tx fee:{' '}
+          {isLoadingFeeEstimate ? (
+            <SpinnerIcon className="inline animate-spin" height="1.2rem" width="1.2rem" />
+          ) : (
+            <>
+              {mintTxFeeEstimation}
```
Sneh1999, line 103: Hmm, I think we should also add `0.001 ETH` that is required to mint an NFT and display the estimated total cost, what are your thoughts on that?

Yuripetusko: is 0.001 a requirement? I couldn't see it in a CustomERC1155 contract, and it's not verified so couldn't check in basescan. I assumed this is a free mint so hence I opted for displaying transaction fee instead of mint price. I will try to get the mint price later, but for some reason I thought I tried and couldn't see it

template/src/pageComponents/mint/ContractDemo.tsx
Yuripetusko: Do you mean something like?

```
{isLoading && (
  <PageLoadingState />
)
```
Or how do you see it? happy to change it if you let me know

template/src/pageComponents/mint/ContractDemo.tsx
Sneh1999: I think you are right, seems like the `ethAmount` here is was incorrectly added and this is actually a free mint. Nice catch!

template/src/pageComponents/mint/ContractDemo.tsx
Sneh1999: I removed the comment, I think this is good as is :)

template/src/pageComponents/mint/ContractDemo.tsx
Yuripetusko: I'll add a comment, so when PaidMint example is added, devs don't forget to add a mint price there

Sneh1999: Awesome! Thank you

template/src/pageComponents/mint/ContractDemo.tsx
Yuripetusko: Thanks. I do think a wrapper can be added, something like <Loader /> where tailwind's animation classes are already added so we don't need to add it every time. But maybe this can be done in a separate pr

Sneh1999: yea, I was thinking that too but yea maybe we can do this in a separate PR. 

CC @arsood @Zizzamia  for later

kyhyco (approved):
LGTM!

Sneh1999 (approved):

template/src/components/icons/SpinnerIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+type Props = React.SVGProps<SVGSVGElement>;
+export function SpinnerIcon(props: Props) {
```
Zizzamia, line 4: Side note, is this something Radix Icons can solve?

I would like avoid add custom icons when Radix that is already imported can solve.

cc @Sneh1999 @Yuripetusko

Yuripetusko: Nope, I looked for the appropriate icon there and it's not there. [lucide-react](https://www.npmjs.com/package/lucide-react) is very popular now for icons, and I use it on some projects, so is react-icons (although this one sometimes has problems with tree-shaking, but maybe they already solved it)

Zizzamia: Great context

PR #357, opened 2024-02-12
fix: notConnected behaviour after moving to wagmi 2.0

Yuripetusko:
**What changed? Why?**

Prior to wagmi 2.0 `useChain` hook was used to determine if correct chain is connected as well as to check if wallet is connected. In wagmi 2.0 the hook was replaced with useAccount and behaviour slightly changed, with a correct way to detect unconnected wallet to check for isConnected property. And `chain` will can still be returned if user haven't connected his account to the dapp

**Notes to reviewers**

**How has it been tested?**

`yarn test`
`manual testing`

### DIFF
--- a/template/src/hooks/contracts.ts
+++ b/template/src/hooks/contracts.ts
@@ -1,4 +1,4 @@
-import { Abi, type Chain } from 'viem';
+import { Abi, Address, type Chain } from 'viem';
 import { baseSepolia } from 'viem/chains';
 import { useAccount } from 'wagmi';
 import BuyMeACoffeeABI from '../contract/BuyMeACoffee';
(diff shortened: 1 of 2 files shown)


template/src/hooks/contracts.ts
```diff
@@ -7,12 +7,12 @@ import SignatureMint721ABI from '../contract/SignatureMint721';
 
 type ContractInstance = {
   chain: Chain;
-  address: `0x${string}`;
+  address: Address;
```
Zizzamia, line 10: Nice

template/src/hooks/contracts.ts
```diff
@@ -28,14 +28,14 @@ type Spec<T extends Abi> = {
  */
 export function generateContractHook<T extends Abi>({ abi, ...spec }: Spec<T>) {
   function useContract(): UseContractReturn<typeof abi> {
-    const { chain } = useAccount();
+    const { chain, isConnected } = useAccount();
```
Zizzamia, line 31: That's neat

Zizzamia (approved):
Love that!!!

PR #368, opened 2024-02-14
fix: use supported chains by env

eragon512:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/template/web/src/providers/OnchainProviders.tsx
+++ b/template/web/src/providers/OnchainProviders.tsx
@@ -1,16 +1,10 @@
 'use client';
 
 import { ReactNode } from 'react';
-import {
-  RainbowKitProvider,
-  connectorsForWallets,
-  darkTheme,
-  lightTheme,
-} from '@rainbow-me/rainbowkit';
-import { metaMaskWallet, rainbowWallet, coinbaseWallet } from '@rainbow-me/rainbowkit/wallets';
+import { RainbowKitProvider, darkTheme, lightTheme } from '@rainbow-me/rainbowkit';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
-import { WagmiProvider, createConfig, http } from 'wagmi';
-import { base, baseSepolia } from 'wagmi/chains';
+import { WagmiProvider } from 'wagmi';
+import { createWagmiConfig } from '../store/createWagmiConfig';
 
 type Props = { children: ReactNode };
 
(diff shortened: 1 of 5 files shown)


Zizzamia:
@eragon512 would you mind add some UnitTests for this, please.

template/web/src/providers/OnchainProviders.tsx
```diff
@@ -41,20 +42,20 @@ const connectors = connectorsForWallets(
...
-    [base.id]: http(),
-  },
-  connectors,
-});
-
 /**
  * TODO Docs ~~~
  */
 function OnchainProviders({ children }: Props) {
+  const wagmiConfig = createConfig({
```
Zizzamia, line 49: Not sure this is the right move.

What happens if the Provider components re-render, I assume the `createConfig` function will be called again.

Zizzamia: I think is best run this outiside the React life-cycle. So that also it starts the execution right away.

eragon512: resolved

template/web/src/store/createWagmiConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+export function createWagmiConfig(projectId: string) {
+  const connectors = connectorsForWallets(
+    [
+      {
+        groupName: 'Recommended Wallet',
+        wallets: [coinbaseWallet],
+      },
+      {
+        groupName: 'Other Wallets',
+        wallets: [rainbowWallet, metaMaskWallet],
```
Sneh1999, line 16: nit: we should add more wallets here

template/web/src/store/supportedChains.ts
```diff
@@ -2,23 +2,23 @@ import { baseSepolia, Chain, base } from 'viem/chains';
...
-export const SUPPORTED_CHAINS = new Map<Environment, Chain[]>([
-  [Environment.localhost, [baseSepolia]],
-  [Environment.development, [baseSepolia]],
-  [Environment.staging, [base]],
-  [Environment.production, [base]],
-]);
+export const SUPPORTED_CHAINS: Record<Environment, [Chain, ...Chain[]]> = {
+  [Environment.localhost]: [baseSepolia],
+  [Environment.development]: [baseSepolia],
+  [Environment.staging]: [base],
```
Sneh1999, line 8: should add `baseSepolia` here as well and also in prod because someone might want to deploy the app only to baseSepolia

template/web/src/store/createWagmiConfig.ts
eragon512: Will take this up as a separate PR to keep each PR atomic and small

template/web/src/store/supportedChains.ts
eragon512: done 👍

Zizzamia (approved):
Niiice

PR #372, opened 2024-02-15
fix: updated buy a coffee contract to include memo paging + text limits

robpolak:
**What changed? Why?**
1. Updated the Buy A Coffee contract.
  1. Added a page and offset to getMemos()
  2. Added caps to message sizes to prevent abuse
1. Updated frontend components.

**Notes to reviewers**
1. Paging still needs to be implemented and we will currently only support 25 messages by default.

**How has it been tested?**
1. Deployed Contract and tested locally.

image


kyhyco (approved):

template/contracts/src/BuyMeACoffee.sol
```diff
@@ -25,7 +25,7 @@ pragma solidity 0.8.23;
  * @dev Memo struct
  */
 struct Memo {
-    uint numCoffees;
+    uint256 numCoffees;
```
Zizzamia, line 28: Out of curiosity, what this change?

template/contracts/src/BuyMeACoffee.sol
```diff
@@ -47,7 +47,14 @@ contract BuyMeACoffee {
     error OnlyOwner();
 
     event BuyMeACoffeeEvent(address indexed buyer, uint256 price);
-    event NewMemo(address indexed userAddress, uint256 time, uint numCoffees, string userName, string twitterHandle, string message);
+    event NewMemo(
+        address indexed userAddress,
+        uint256 time,
+        uint256 numCoffees,
+        string userName,
```
Zizzamia, line 54: I kind of think we should stop ask the `userName` and the `twitterHandle` and instead use ENS data for this.

template/contracts/src/BuyMeACoffee.sol
```diff
@@ -143,8 +159,27 @@ contract BuyMeACoffee {
     /**
      * @dev Function to get the memos
      */
-    function getMemos() public view returns (Memo[] memory) {
-        return memos;
+    function getMemos(uint256 index, uint256 size) public view returns (Memo[] memory) {
```
Zizzamia, line 162: @Sneh1999 @awilliams1-cb curios to hear thoughts on this approach.

template/contracts/test/BuyMeACoffee.t.sol
```diff
@@ -3,45 +3,91 @@ pragma solidity 0.8.23;
...
         assertEq(memo.message, "new message");
     }
 
     function testWithdrawTips() public {
+        buyMeACoffee.buyCoffee{value: 0.0001 ether}(numCoffees, userName, twitterHandle, message);
         buyMeACoffee.withdrawTips();
         assertEq(address(buyMeACoffee).balance, 0);
     }
 
+    function testPaging() public {
```
Zizzamia, line 49: Niiiiice

template/web/src/hooks/useOnchainCoffeeMemos.ts
```diff
@@ -16,6 +16,7 @@ function useOnchainCoffeeMemos() {
     address: contract.status === 'ready' ? contract.address : undefined,
     abi: contract.abi,
     functionName: 'getMemos',
+    args: [BigInt(0), BigInt(25)], // TODO : Implement Paging
```
Zizzamia, line 19: Oh very interesting

template/contracts/src/BuyMeACoffee.sol
Sneh1999: I dont think paginating inside the smart contract is a general pattern used, @awilliams1-cb feel free to correct me here though

template/contracts/src/BuyMeACoffee.sol
awilliams1-cb: `uint` is [an alias](https://docs.soliditylang.org/en/v0.8.23/types.html#integers) for `uint256`. In general, using the more explicit `uint256` is preferred, and then that should be used consistently across the contract.

robpolak: This was done by the "forge fmt" command.. we should ensure that is being run on checkin.

template/contracts/src/BuyMeACoffee.sol
robpolak: Typically, I would just iterate event logs, but that is not a very efficient mechanism across a large blockspan (I think most providers will prevent you going back too far, 1000 blocks maybe?).  Agree in a production environment we would be indexing these events and putting them in a low-latency data store.

awilliams1-cb: It's definitely more common to have pagination for non-view functions. In general, I would say the way people handle data like this is to watch for the corresponding event, then save that data as it streams in (or use a subgraph / other data aggregation tool).

For our use case I think it's fine to paginate this view function though

Sneh1999: makes sense, I remember @Zizzamia  wanted to ad to add `the graph` to boat, maybe it can be used here. We could also merge this PR in the meantime and add graph later whatever you guys think is the best

template/contracts/src/BuyMeACoffee.sol
```diff
@@ -143,8 +159,27 @@ contract BuyMeACoffee {
...
+        if (index >= memos.length) {
```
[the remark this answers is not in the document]
robpolak: fixed.. just returns an empty array

template/contracts/src/BuyMeACoffee.sol
Zizzamia: 100% with you on this

template/contracts/src/BuyMeACoffee.sol
Zizzamia: i am ok to merge this PR now, and than as follow up we start working on using `the graph`.

PR #376, opened 2024-02-16
chore: add more Web3 wallets

eragon512:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Manual

Screenshot 2024-02-17 at 12 56 35 AM

### DIFF
--- a/template/web/src/store/createWagmiConfig.ts
+++ b/template/web/src/store/createWagmiConfig.ts
@@ -1,5 +1,13 @@
 import { connectorsForWallets } from '@rainbow-me/rainbowkit';
-import { metaMaskWallet, rainbowWallet, coinbaseWallet } from '@rainbow-me/rainbowkit/wallets';
+import {
+  metaMaskWallet,
+  rainbowWallet,
+  coinbaseWallet,
+  trustWallet,
+  ledgerWallet,
+  braveWallet,
+  walletConnectWallet,
+} from '@rainbow-me/rainbowkit/wallets';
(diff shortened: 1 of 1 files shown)


template/web/src/store/createWagmiConfig.ts
```diff
@@ -13,7 +21,14 @@ export function createWagmiConfig(projectId: string) {
       },
       {
         groupName: 'Other Wallets',
-        wallets: [rainbowWallet, metaMaskWallet],
+        wallets: [
+          rainbowWallet,
+          metaMaskWallet,
+          trustWallet,
```
Sneh1999, line 27: @eragon512 have you tested with these wallets?

template/web/src/store/createWagmiConfig.ts
```diff
@@ -13,7 +21,14 @@ export function createWagmiConfig(projectId: string) {
...
+          braveWallet,
+          ledgerWallet,
+          walletConnectWallet,
```
Zizzamia, line 30: Wallet Connect adds trash in the HTML, let's not add it. We can change our mind after they clean up their library.

template/web/src/store/createWagmiConfig.ts
Zizzamia: And for each Wallet, I would like we triple check the HTML is trashed.

eragon512:
Not able to test the other wallets at all, closing this PR for now

PR #383, opened 2024-02-17
[chore] delete unused package.json

kyhyco:
**What changed? Why?**
Delete unused `package.json` file

**Notes to reviewers**

**How has it been tested?**
N/A

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -1,8 +1,0 @@
-{
-  "name": "package_name",
-  "description": "",
-  "version": "0.0.1",
-  "keywords": [],
-  "author": "",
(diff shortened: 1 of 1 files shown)


template/package.json
```diff
@@ -1,8 +0,0 @@
-{
```
Zizzamia, line 1: Without the package there is an error during the CLI first run.

Zizzamia: Maybe can you confirm why that happens?

Also we should probably have a README to welcome folks

Zizzamia:
I belive we deleted this

PR #390, opened 2024-02-19
[feature] Improved CLI experience

kyhyco:
**What changed? Why?**
- Handle `.env.local` and `.env` file creation
- Use `piccolors` instead of `chalk` (saving ~100kb)
- Use `@clack/prompts` instead of `inquirer`
- Update `template/web/.env.local.default` to `template/web/.env.local`
- Overall faster and snappier experience

https://github.com/coinbase/build-onchain-apps/assets/132851666/e169f1fa-12af-4b81-a720-0f2cfff3c8d3

**Notes to reviewers**

**How has it been tested?**
locally


package.json
```diff
@@ -4,12 +4,11 @@
   "repository": "https://github.com/coinbase/build-onchain-apps.git",
   "license": "Apache-2.0",
   "dependencies": {
-    "chalk": "^4.1.2",
+    "@clack/prompts": "^0.7.0",
     "commander": "^11.1.0",
     "figlet": "^1.7.0",
-    "inquirer": "^8.2.5",
-    "rimraf": "^5.0.5",
-    "tslib": "^2.3.0"
```
kyhyco, line 12: don't think `tslib` is used any where

src/create/note.ts
```diff
@@ -0,0 +1,71 @@
+import * as color from 'picocolors';
```
kyhyco, line 1: Improving visibility of message in `note` from `@clack/prompts`.

`@clack/prompts` dims the `note` message color making it hard to see.

package.json
kyhyco: it is being used by GitHub Actions...

Zizzamia (approved):

PR #394, opened 2024-02-19
[feature] module installs

kyhyco:
**What changed? Why?**
Allow users to decide to include demo apps

- move test files to `test` folders
- move `src/global.css` to `app/global.css` which follows the NextJS convention
- move app specific contract files moved to `app/<name>/_contracts`

**Notes to reviewers**

**How has it been tested?**
locally


template/web/src/hooks/MockABI.ts
```diff
@@ -0,0 +1,127 @@
+/**
+ * This ABI is trimmed down to just the functions we expect to call for the
+ * sake of minimizing bytes downloaded.
+ */
+const abi = [
```
Zizzamia, line 5: I don't follow, why are we faking this?

template/web/app/(demo)/mint/_contracts/useCustom1155Contract.test.ts
```diff
@@ -0,0 +1,15 @@
+import { baseSepolia } from 'viem/chains';
+import Custom1155ABI from './Custom1155ABI';
+import { useCustom1155Contract } from './useCustom1155Contract';
+
```
Zizzamia, line 4: Let's avoid this pattern `template/web/app/(demo)` where there is a `demo` folder.

Zizzamia: each page that comes with BOAT is already a full functioning demo experience.

template/web/src/hooks/MockABI.ts
Zizzamia: We should use real ABI for real experiences.

src/create/getUserInput.ts
```diff
@@ -24,6 +24,12 @@ export async function getUserInput() {
           },
         });
       },
+      setupModules: () => {
+        return prompts.confirm({
+          message: 'Install demo apps',
```
Zizzamia, line 29: I don't like the word `demo` or `apps`.

Those are experiences part of one app.

And App can have multiple experiences.

template/web/src/hooks/MockABI.ts
Zizzamia, line 5: let's move all tests into the `test` folder a `src/hooks/test` so also the Mock goes there.

src/create/getUserInput.ts
kyhyco: any suggestions on the wording?

kyhyco: Actually how about `experiences`? It would be nice to create a grouping for the BOAT experiences. Something like `(experiences)/_layout` (containing header, footer, etc) would be useful.

I'd like to make `src/*` mostly to contain app agnostic code and move app specific components closer to `app/(experiences)`.

Zizzamia: Oh I see, you were never part of the sync meetings we did before. We always called them experiences. :)

Zizzamia (approved):
That's pretty neat

Let's go ahead finish this PR before other folks start coding on it tomorrow

src/create/getUserInput.ts
```diff
@@ -26,7 +26,7 @@ export async function getUserInput() {
       },
       setupModules: () => {
         return prompts.confirm({
-          message: 'Install demo apps',
+          message: 'Include dApp experiences',
```
Zizzamia, line 29: Onchain App

Zizzamia (approved):

PR #395, opened 2024-02-19
[chore] update gif

kyhyco:
**What changed? Why?**
new gif with reduced height

**Notes to reviewers**

**How has it been tested?**


Zizzamia (approved):
That's pretty slick

PR #396, opened 2024-02-20
[test] test commit

kyhyco:
**What changed? Why?**
checking user is correctly registered in the commit

**Notes to reviewers**

**How has it been tested?**


kyhyco:
looks like it works

PR #407, opened 2024-02-21
feat: read onchain data without connected wallet

eragon512:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Manual


kyhyco (approved):
verified locally and it works great!

PR #416, opened 2024-02-22
feat: show NFT image without connected wallet

eragon512:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Manual

Screenshot 2024-02-22 at 2 40 58 PM


Zizzamia (approved):

Zizzamia:
nit. we should probably show some alert to tell folks to connect the wallet to mint

PR #418, opened 2024-02-22
enhancement: Paymaster Bundler Design Review Feedback Addressed

arsood:
**What changed? Why?**

- Added footer link
- Added Google Font - Roboto Mono and used it
- Add item amount
- Refactor Vault to use Item common component

**How has it been tested?**

- Locally

### DIFF
--- a/template/web/app/global.css
+++ b/template/web/app/global.css
@@ -1,6 +1,3 @@
-/* Google Fonts - Roboto Mono */
-@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono:ital,wght@0,100..700;1,100..700&display=swap');
-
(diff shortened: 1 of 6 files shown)


template/web/app/global.css
```diff
@@ -1,3 +1,6 @@
+/* Google Fonts - Roboto Mono */
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono:ital,wght@0,100..700;1,100..700&display=swap');
```
Zizzamia, line 2: Why this?

I thought we already have this font download.

arsood: @Zizzamia oh do we? I didn't see it in the codebase.

arsood: Yes you're right, I'll fix

kyhyco (approved):

PR #420, opened 2024-02-22
fix: Cleanup UI bugs reported in bug bash Feb, 2024

Sneh1999:
**What changed? Why?**
Fixing some UI bugs before ETH Denver which were reported in the bug bash

**Notes to reviewers**

**How has it been tested?**
Note: I have deleted the Profile.tsx code because I wanted to minimize the amount of unwanted code we ship with our package, we can add it later once we have an idea on how we want to dynamically create this view


Zizzamia (approved):
Dope

PR #430, opened 2024-02-23
Add WIP banner for Paymaster page

Sneh1999:
**What changed? Why?**
Adding a warning banner for all future use cases

image

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/template/web/src/components/layout/banner/banner.tsx
+++ b/template/web/src/components/layout/banner/banner.tsx
@@ -11,17 +11,12 @@ type BannerProps = {
 export default function Banner({ pageName, pageUrl, wip }: BannerProps) {
   return (
     <section
-      className={
-        wip
-          ? clsx(
-              'flex flex-col items-center justify-between gap-6 p-6 md:flex-row md:gap-0',
-              'rounded-lg border border-zinc-400 border-opacity-20 bg-boat-color-yellow-60 text-black',
-            )
-          : clsx(
-              'flex flex-col items-center justify-between gap-6 p-6 md:flex-row md:gap-0',
-              'rounded-lg border border-zinc-400 border-opacity-20  bg-neutral-900 text-white',
-            )
-      }
+      className={clsx(
+        'flex flex-col items-center justify-between gap-6 p-6 md:flex-row md:gap-0',
+        `rounded-lg border border-zinc-400 border-opacity-10  ${
+          wip ? `bg-boat-color-yellow-60 text-black` : `bg-neutral-900 text-white`
+        }`,
+      )}
     >
       <div className="flex items-start justify-start gap-2 md:gap-6">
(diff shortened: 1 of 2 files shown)


Zizzamia:
We should probably ask Rose to get a more balanced yellow.

kyhyco (approved):

template/web/src/components/layout/banner/banner.tsx
```diff
@@ -5,15 +5,23 @@ import NextLink from 'next/link';
...
-export default function Banner({ pageName, pageUrl }: BannerProps) {
+export default function Banner({ pageName, pageUrl, wip }: BannerProps) {
   return (
     <section
-      className={clsx(
-        'flex flex-col items-center justify-between gap-6 p-6 md:flex-row md:gap-0',
-        'rounded-lg border border-zinc-400 border-opacity-20 bg-neutral-900',
-      )}
+      className={
+        wip
```
arsood, line 15: Instead of this condition switching between these large blocks we can do this directly inside clsx because there are only slight differences between the two. This will allow us to use only one clsx call and simplify this functionality.

template/web/src/components/layout/banner/banner.tsx
```diff
@@ -26,14 +34,22 @@ export default function Banner({ pageName, pageUrl }: BannerProps) {
...
             Step into the {pageName} experience.
           </h1>
-          <div className="font-inter text-base font-normal leading-normal text-white">
-            Take a dive into our fully functional demo page. Explore how the Smart Contract works,
-            connect with it using Wagmi, and master a seamless user experience with React and
-            Tailwind.
-          </div>
+          {wip ? (
+            <div className="font-inter text-base font-normal leading-normal ">
+              Just a quick heads up: this experience is a work in progress! 👀 Keep an eye out for
```
arsood, line 42: Great!

template/web/src/components/layout/banner/banner.tsx
```diff
@@ -5,14 +5,17 @@ import NextLink from 'next/link';
...
 };
 
-export default function Banner({ pageName, pageUrl }: BannerProps) {
+export default function Banner({ pageName, pageUrl, wip }: BannerProps) {
   return (
     <section
       className={clsx(
         'flex flex-col items-center justify-between gap-6 p-6 md:flex-row md:gap-0',
-        'rounded-lg border border-zinc-400 border-opacity-20 bg-neutral-900',
+        `rounded-lg border border-zinc-400 border-opacity-10  ${
```
arsood, line 16: nit: Don't need to use a template string literal here. With clsx we can just use a standard ternary with string values.

Sneh1999: Done!

PR #431, opened 2024-02-23
feat: Paymaster Bundler - Experience

arsood:
**What changed? Why?**

- Add integration with gameplay functionality
- Button and game states
- Restart functionality
- Animation with React Spring

**How has it been tested?**

- Locally


template/web/app/paymaster-bundler/constants.ts
```diff
@@ -0,0 +1,29 @@
+export const rpcUrl =
+  'https://api.developer.coinbase.com/rpc/v1/base-sepolia/Nz9nSdr8uIY7HhIi2zRrJFgISslFF2In';
+export const paymasterUrl =
+  'https://api.developer.coinbase.com/rpc/v1/base-sepolia/Nz9nSdr8uIY7HhIi2zRrJFgISslFF2In';
```
kyhyco, line 4: Should this be passed via `process.env.NEXT_PUBLIC_RPC_URL` instead?

template/web/app/paymaster-bundler/constants.ts
```diff
@@ -0,0 +1,29 @@
+export const rpcUrl =
+  'https://api.developer.coinbase.com/rpc/v1/base-sepolia/Nz9nSdr8uIY7HhIi2zRrJFgISslFF2In';
```
kyhyco, line 2: Should this be passed via process.env.RPC_URL instead?

template/web/app/paymaster-bundler/constants.ts
kyhyco, line 4: Does this need to be API key from Pimlico?

template/web/app/paymaster-bundler/_components/GamePlay.tsx
```diff
@@ -1,31 +1,180 @@
...
+    // 25% chance of getting 0
+    randomNumber = 0;
+  } else {
+    // 75% chance of getting a number between 1 and 3
+    randomNumber = Math.floor(randomValue * 3) + 1;
+  }
+  return randomNumber;
+};
+
+export default function GamePlay({ setOwnedTokens, smartAccount, client }: GameplayProps) {
```
kyhyco, line 37: The current GamePlay only works when logging in with an email account but not with Coinbase Wallet. Is that intentional? Is the wallet flow suppose to work too in this PR?

template/web/app/paymaster-bundler/constants.ts
0xAlec: done in new commit

template/web/app/paymaster-bundler/constants.ts
0xAlec: done in new commit, we're using the Cloud Paymaster here

template/web/app/paymaster-bundler/_components/GamePlay.tsx
0xAlec: removed, wasn't intended to allow users to use their wallet

template/web/app/paymaster-bundler/_components/abi.tsx
```diff
@@ -0,0 +1,633 @@
+export const nftAbi = [
```
robpolak, line 1: What contract is this?  If this is a generic erc721/1155 let's call that out.

template/web/.env.local.example
```diff
@@ -1,3 +1,6 @@
 NEXT_PUBLIC_GOOGLE_ANALYTICS_ID=
 NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID="GET_ID_FROM_WALLET_CONNET" # See https://cloud.walletconnect.com
+NEXT_PUBLIC_RPC_URL="GET_FROM_COINBASE_CLOUD"
```
robpolak, line 3: I worry this example won't work out-of-the-box for new users.  Is there a way for us to support this at a project level on Sepolia without users needing to create accounts?

template/web/app/paymaster-bundler/_components/PaymasterBundlerApp.tsx
```diff
@@ -0,0 +1,30 @@
+import { useCallback } from 'react';
+import { PrivyProvider, User } from '@privy-io/react-auth';
+import { privyID } from '../constants';
+import PaymasterBundlerDemo from './PaymasterBundlerDemo';
+
+export default function PaymasterBundlerApp() {
+  const handleLogin = useCallback((user: User) => {
+    console.log(`Successfully authenticated with Privy! Your User ID: ${user.id}`);
```
robpolak, line 8: Do we need to log this?  Will the user need their ID?

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
```diff
@@ -0,0 +1,60 @@
...
+      args: [Number(tokenID)],
+    });
+
+    tokens.push(tokenJSONLink as string);
+  }
+
+  const fetchOps = [] as Promise<unknown>[];
+  const tokenJSONs = [] as NFTType[];
+
+  tokens.forEach((token) => {
```
robpolak, line 42: This is crazy inefficient for large collections, it will end up locking up the UI and sending a large amount of requests to the RPC provider.  I know this is not a simple solution, however, we should probably use a data aggregator here. cc: @Zizzamia .. WDYT about integrating with an API vendor for these type of aggregated data lookups?

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
```diff
@@ -0,0 +1,60 @@
...
+    abi: nftAbi,
+    functionName: 'balanceOf',
+    args: [address],
+  });
+
+  // Get the token IDs and metadata by token ID
+  var tokens = [] as string[];
+
+  for (let i = 0; i < Number(numTokens); i++) {
+    const tokenID = await client.readContract({
```
robpolak, line 22: This is crazy inefficient for users who own a large amount of tokens, it will end up locking up the UI and sending a large amount of requests to the RPC provider. I know this is not a simple solution, however, we should probably use a data aggregator here. cc: @Zizzamia .. WDYT about integrating with an API vendor for these type of aggregated data lookups?

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
```diff
@@ -0,0 +1,60 @@
...
+      address: '0x66519FCAee1Ed65bc9e0aCc25cCD900668D3eD49',
+      abi: nftAbi,
+      functionName: 'tokenOfOwnerByIndex',
```
robpolak, line 25: Worth noting this approach will only work with contracts that support IERC721Enumerable.. might confuse users why this won't work with a standard ERC721 contract.

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { nftAbi } from '../_components/abi';
+import { NFTType } from '../types';
+
+export default async function fetchNFTs(smartAccount: SmartAccountClient, client: PublicClient) {
+  // Get # of NFTs owned by address
+  const address = smartAccount.account?.address;
+
+  const numTokens = await client.readContract({
+    address: '0x66519FCAee1Ed65bc9e0aCc25cCD900668D3eD49',
```
Zizzamia, line 12: In the follow up PR let's rewrite this, and use `generateContractHook`. cc @arsood

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  const fetchOps = [] as Promise<unknown>[];
+  const tokenJSONs = [] as NFTType[];
+
+  tokens.forEach((token) => {
+    fetchOps.push(
+      (async () => {
+        try {
+          const tokenResponse = await fetch(token);
```
robpolak, line 46: Per above, this will be inefficient for users with a large amount of tokens.  This will fail for IPFS URIs as well.

template/web/app/paymaster-bundler/constants.ts
```diff
@@ -0,0 +1,28 @@
+export const rpcUrl = process.env.NEXT_PUBLIC_RPC_URL;
+export const paymasterUrl = process.env.NEXT_PUBLIC_PAYMASTER_URL;
+export const privyID = 'clsxnba7r00tragoglekx9or7';
```
robpolak, line 3: This safe to checkin?

template/web/app/paymaster-bundler/_components/PaymasterBundlerApp.tsx
arsood: Good point, removed.

template/web/app/paymaster-bundler/constants.ts
arsood: No, removed and @0xAlec will cycle keys.

template/web/app/paymaster-bundler/_components/GamePlay.tsx
```diff
@@ -1,31 +1,180 @@
...
+      const randomNumber = getRandomNumber();
+
+      const data = encodeFunctionData({
+        abi: nftAbi,
+        functionName: 'mintTo',
+        args: [smartAccount.account?.address, randomNumber],
+      });
+
+      try {
+        await smartAccount.sendTransaction({
```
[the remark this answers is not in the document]
0xAlec: it is not, viem docs suggest to use a 3rd party AA SDK provider

happy to replace with viem natively if they decide to support AA in the future

kyhyco (approved):

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
Yuripetusko: keep in mind that viem has `readContracts` (plural) function too, it can even support multicall3. I'm happy to create a mini pr into this one if you want

Yuripetusko: I also agree about the danger, as you never know how many tokens user has. maybe it's good as an example to have either pagination or limit set.

Regarding "api vendor", IMO it's good to show no-indexer example to devs, so people don't learn to be so reliant on indexers and centralised apis. But in real world, indexer/api is the most efficient way probably, so perhaps it's good to create both examples (or hooks). Fetching from contract with pagination or a hard limit, and another one from alchemy/reservoir etc..

template/web/app/paymaster-bundler/_components/Vault.tsx
```diff
@@ -11,15 +17,17 @@ export default function Vault() {
...
-        <Item
-          src="/account_abstraction/wing_suit.png"
-          altText="Wing Suit"
-          active={false}
-          amount={4}
-        />
+        {ALL_ITEMS.map((item) => {
+          return (
+            <Item
+              key={item.name}
```
Yuripetusko, line 23: perhaps name can be duplicated, in this case maybe appending it with token id is good? `key={`${item.name}-${item.id}`}`

template/web/app/paymaster-bundler/_utils/fetchNFTs.ts
Yuripetusko: sorry ignore me. just noticed it's already merged

PR #439, opened 2024-02-24
use react-query in fetching hooks; improve ipfs utils 

### LINKED ISSUE #356: Feature Request: Use react-query and/or wagmi instead of custom fetch hooks for datafetching
### Describe the solution you'd like

react-query is already a dependency and it is also part of wagmi hooks. It handles caching very well, you can pass expiration, cache time to each hook and also use custom storage if needed.

I noticed 3 hooks that use custom solutions where I beleive react-query or wagmi would work better

- `useEnsName `
- `ensAvatar `
- `useCollectionMetadata`

### Describe alternatives you've considered.

`useCollectionMetadata` should use react query to use all the benefits of react-query caching and have consistent api with other hooks. Both `useEnsName` and `ensAvatar` can use wagmi's hooks directly. Alternatively a custom react-query hook similar to:

```
/**
 * A hook to get ENS avatar
 * @param ensAddressOrName - either EVM address or an ens name like bitfalls.eth
 * @param chainId - defaults to 1 for ethereum as generally ENS is there only
 * @param skip - used to conditionally skip the hook call
 */
export const useGetEnsIdentity = ({
  ensAddressOrName,
  chainId = ENS_CHAIN_ID,
  skip = false,
}: {
  ensAddressOrName?: string;
  chainId?: number;
  skip?: boolean;
}) => {
  return useQuery<string | null>({
    queryKey: ['useGetEnsAvatar', ensAddressOrName, chainId],
    queryFn: async () => {
      const ensName =
        ensAddressOrName && getIsValidEthereumAddress(ensAddressOrName)
          ? await getEnsName(wagmiConfig, { address: ensAddressOrName, chainId })
          : ensAddressOrName;
      const ensResolver = ensName
        ? await getEnsResolver(wagmiConfig, { chainId, name: ensName })
        : null;

      const publicClient = getPublicClient(wagmiConfig, {
        chainId,
      });

      if (!ensResolver || !ensName) {
        return null;
      }

      const avatar = await publicClient.getEnsAvatar({ name: normalize(ensName) });

      return avatar;
    },
    gcTime: CACHE_TIME,
    enabled: !skip,
    refetchOnWindowFocus: false,
  });
};
```

Yuripetusko:
**What changed? Why?**
Closes #356 

- Use @tanstack/react-query in fetching hooks (useCollectionMetadata and useCurrentBlockNumber)
  
- On useCollectionMetadata:
  - Allow to pass chainId explicitly
  - Improve types
  - Don't assume that all metadata fields are present anymore
  
- Since now the type of JsonMetadata is more correct, and all fields can be undefined (saw too many nft metadatas like this), displaying fallback image if image is not present and fallback to collection address if name is not present (See screenshots below)

**Notes to reviewers**

`useCurrentBlockNumber` doesn't make sense to me and I think it needs to be removed, as `wagmi` already has a hook like that. It's also not used anywhere, so no way to test except for unit testing

**How has it been tested?**

- Run `yarn test`
- Did visual testing

----

Fallbacks screenshots when image or name are not present in metadata:

Screenshot 2024-02-25 at 00 16 22
Screenshot 2024-02-25 at 00 09 58


Zizzamia:
Ciao, first of all, thank you for the help.

Would you mind split this PR in smaller PRs? It can help us evalute them faster.

Yuripetusko:
> Ciao, first of all, thank you for the help.
> 
> Would you mind split this PR in smaller PRs for each fix or improvement? It can help us evalute them faster.

Sure I can try, although these are a bit dependant on each other. But I guess the logical parts are:

1. 2 hooks that now use react-query 
2. ipfs util changes and addition of `is-ipfs`
3. Fallback image component and collection name fallback

The ipfs util is easiest to split, but if I split fallback image, then there will be merge conflicts, since `image` and `name` fields of metadata can be undefined, and new types are part of the `useCollectionMetadata` hook, this will be weird to do without the other

Will splitting just ipfs util changes be enough in this case?

Yuripetusko:
> > Ciao, first of all, thank you for the help.
> > Would you mind split this PR in smaller PRs for each fix or improvement? It can help us evalute them faster.
> 
> Sure I can try, although these are a bit dependant on each other. But I guess the logical parts are:
> 
> 1. 2 hooks that now use react-query
> 2. ipfs util changes and addition of `is-ipfs`
> 3. Fallback image component and collection name fallback
> 
> The ipfs util is easiest to split, but if I split fallback image, then there will be merge conflicts, since `image` and `name` fields of metadata can be undefined, and new types are part of the `useCollectionMetadata` hook, this will be weird to do without the other
> 
> Will splitting just ipfs util changes be enough in this case?

Ok moved ipfs util changes to separate PR now

Zizzamia:
@Yuripetusko thank you so much, just FYI the team is onsite at ETH Denver for the whole week, so we are going to be a bit slow in reviewing and merging PRs this week.

Yuripetusko:
> @Yuripetusko thank you so much, just FYI the team is onsite at ETH Denver for the whole week, so we are going to be a bit slow in reviewing and merging PRs this week.
> 

Have fun there. I'll see the notification whenever the team does the review.

Zizzamia:
@cnasc @robpolak @Sneh1999  when you have a minute, can you help review this one.

template/web/src/components/FallbackImage/FallbackImage.module.css
```diff
@@ -0,0 +1,4 @@
+.FallbackImageIcon {
+  width: clamp(0%, 50%, 100%);
```
snehkoul-cb, line 2: Can we use Tailwind here if possible?

template/web/src/hooks/useBlockNumber.ts
```diff
@@ -1,56 +1,41 @@
...
+type Props = {
+  refetchInterval?: number;
+  enabled?: boolean;
+  chainId?: Chain['id'];
+};
 
-  useEffect(() => {
-    let intervalId: NodeJS.Timeout;
+//TODO: consider using wagmi's `useBlockNumber` instead
+const useCurrentBlockNumber = (props?: Props) => {
```
snehkoul-cb, line 16: Whats the reasoning behind getting the chainId from the arguments if present, why not always take the one connected to the wallet?

template/web/src/hooks/useCollectionMetadata.ts
```diff
@@ -1,115 +1,145 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useQuery } from '@tanstack/react-query';
 import { Abi, Address } from 'abitype';
-import { useReadContract } from 'wagmi';
+import { Chain } from 'viem/chains';
+import { useAccount, usePublicClient } from 'wagmi';
 import { ipfsToHTTP } from '@/utils/ipfs';
 
+// Minimal ABI with contractURI view function
+const AbiWithContractURI = [
```
snehkoul-cb, line 8: can we put the ABI's in files similar to this: https://github.com/coinbase/build-onchain-apps/blob/4d9b38cfefb1dceff36aee0a229168a9a01971b4/template/web/app/mint/_contracts/Custom1155ABI.ts#L2?

template/web/src/hooks/useCollectionMetadata.ts
```diff
@@ -1,115 +1,145 @@
...
+  uri: 'uri',
+  contractURI: 'contractURI',
+} as const;
+
+const getIsAbiWithContractURIViewFunction = (abi: Abi) =>
+  !!abi.find((el) => el.type === 'function' && el.name === uriFunctionType.contractURI);
+const getIsAbiWithUriViewFunction = (abi: Abi) =>
+  !!abi.find((el) => el.type === 'function' && el.name === uriFunctionType.uri);
+
+const isValidAbi = (abi: Abi) =>
```
snehkoul-cb, line 42: I think lets move these functions to utils because they can be used across multiple files

template/web/src/hooks/useBlockNumber.ts
Yuripetusko: Maybe it's a personal preference, but I think it's a bad UX where user needs to switch chain to view your dapp. This is more important for write operation of course, but reads as well. For example you have an NFT dapp that supports NFTs from multiple chains, why would user need to switch chains every time he opens an NFT details page of a different chain from the one connected in a wallet, just for a page to read some contract states.

Perhaps for this particular hook it is not as needed, but I think it's a good practice and ideally all wagmi hook wrappers should always have an option to pass chainId explicitly

snehkoul-cb: Makes sense. I think this might cause confusion to people who are reading this for the first time on why the chainId is added as a prop for this particular hook and not for all the other read hooks. Just so that we maintain a pattern, we should maybe not add this right now and later we can do a more bigger cleanup where we can add chainId to all read/write hooks. Thoughts? 

CC @Zizzamia @kyhyco @cnasc feel free to add your thoughts as well

Yuripetusko: ok makes sense. I can create a discussion or an issue instead

Yuripetusko: btw, I added a comment here also as I think this hook is completly redundant and it's better to just use wagmi's hook

template/web/src/hooks/useCollectionMetadata.ts
Yuripetusko: hmm I spent a bit of time on moving these abis as well as validation utils as a separate utils and it doesn't feel right to me I'll be honest. 

These are not really abis, I should probably rename them to `AbiFunctionContractURI` and `AbiFunctionURI`. As these are more like fragments of the ABIs, and shouldn't be used as abis by mistake.
I kind of like all of this code to be self-contained within the same, especially since it has very specific and somewhat hacky way to check which function to use on the abi (contractURI vs uri). 

I just pushed a change where I move the action away from the hook at least. Let me know if this is better

template/web/src/components/FallbackImage/FallbackImage.module.css
Yuripetusko: Just checked and `Clamp` utility is still not implemented https://github.com/tailwindlabs/tailwindcss/discussions/10647

template/web/src/hooks/useBlockNumber.ts
cnasc: I would agree that in practice people should use the `wagmi` hook. If I remember correctly, this hook was intended only as a beginner example to show some code using `wagmi` in action. This library has evolved quite a bit since when this was written and I don't think it serves the purpose it once did. IMO would +1 on removing it, interested in hearing from the others

template/web/src/components/FallbackImage/FallbackImage.module.css
snehkoul-cb: Do we need clamp or can we use something else?

Yuripetusko: Good question. I used it to have image-fallback svg stretch with the size of the container. I'll see if I can find a simpler alternative

template/web/src/hooks/useCollectionMetadata.ts
Yuripetusko: answered in another comment regarding splitting these away from the hook

snehkoul-cb (approved):
LGTM!

template/web/src/hooks/useBlockNumber.ts
Zizzamia: 3, 2, 1, and it's gone
https://github.com/coinbase/build-onchain-apps/pull/460

Zizzamia:
Ok, sorry for the delay in reviewing this.

I see we are cooking a lot of things in this PR.

> useCurrentBlockNumber doesn't make sense to me and I think it needs to be removed, as wagmi already has a hook like that. It's also not used anywhere, so no way to test except for unit testing

On this one, you right, just got rid of it https://github.com/coinbase/build-onchain-apps/pull/460 .
Let's rebase your PR with `main`.

template/web/app/mint/_components/ContractDemo.tsx
```diff
@@ -71,14 +77,19 @@ export default function MintContractDemo() {
   return (
     <div className="gap-16 lg:flex">
       <div className="w-full flex-shrink-0 flex-grow lg:max-w-[400px] xl:max-w-[600px]">
-        <NextImage
-          src={imageAddress}
-          altText={collectionName}
-          className="block w-full rounded-2xl"
-        />
+        {collectionImageUrl ? (
+          <NextImage
```
Zizzamia, line 81: yeah, more I read this, and more i keep thinking OnchainKit should have a NftMedia component.

template/web/app/mint/_components/ContractDemo.tsx
```diff
@@ -71,14 +77,19 @@ export default function MintContractDemo() {
...
+            className="block w-full rounded-2xl"
+          />
+        ) : (
+          <FallbackImage />
+        )}
       </div>
+
       <div className="flex-shrink-1 mt-10 w-full flex-grow-0 lg:mt-0">
-        <h1 className="text-4xl font-bold">{collectionName}</h1>
+        <h1 className="text-4xl font-bold">{collectionNameOrAddress}</h1>
```
Zizzamia, line 92: Also here, like have a NFTName, component in Onchainkit.

template/web/src/actions/get-collection-metadata.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * There is some differences in URI standards between ERC721 and 1155, handle those in this component.
+ */
+const uriFunctionType = {
+  uri: 'uri',
+  contractURI: 'contractURI',
+} as const;
+
+const getIsAbiWithContractURIViewFunction = (abi: Abi) =>
```
Zizzamia, line 35: Sorry, can we write this code in a more verbose way.

It's a bit hard to follow when it's all written in one solo line of command.

Same in other places in this file.

template/web/src/actions/get-collection-metadata.ts
```diff
@@ -0,0 +1,125 @@
...
+  ) as unknown as typeof AbiFunctionContractURI;
+const getIsAbiWithUriViewFunction = (abi: Abi) =>
+  !!abi.find(
+    (el) => el.type === 'function' && el.name === uriFunctionType.uri,
+  ) as unknown as typeof AbiFunctionWithUri;
+
+const isValidAbi = (abi: Abi) =>
+  getIsAbiWithContractURIViewFunction(abi) || getIsAbiWithUriViewFunction(abi);
+
+export type JsonMetadata = {
```
Zizzamia, line 47: This file has a lot of things going on. Also, I wonder if a lot of this code should just live in OnchainKit.

@Yuripetusko what do you think?

Yuripetusko: Yes @snehkoul-cb noted the same, and rightly so. However when I tried to decouple this I found that each util depends on each other too much. It mostly comes due to the fact that this is trying to get metdata uri from 2 separate functions with different signatures. And typescript doesn't have a way to create type guards for arrays like ABI, so we can't have a union type that it would understand when passing to viem/wagmi. So there's a lot of weird code because of this.

I wonder if this should just be split into separate functions and users would have to use one or the other depending on type of NFT contract (one with `uri` and another with `contractURI` 🤔)

Re: moving this to onchainkit. Yes I think absolutly + possibly the ipfs utils changes I did in another PR here?

Yuripetusko: Yeah the more I think about this the more it makes sense. Most of the code in this file is trying to guess what kind of ABI it is, and then depending on the type, passing different params to viem readContract function. And it's not very easy to move this code into utils because of typescript limitations. So I wonder if instead of trying to make a universal function that checks what kind of abi it is, we just create 2 hooks `useContractUriMetadata` and `useUriMetadata` or something along the lines

Yuripetusko: or `useErc1155Metadata` (with optional tokenId argument) and `useNftCollectionMetadata` (purely fro `contractURI`)

Yuripetusko: ok I think this is the way to go actually, I'll try to make these changes

Yuripetusko: Ok @Zizzamia hopefully it's more readable now. But I should also probably move this to OnChainkit soon

kyhyco (approved):

Zizzamia (approved):
Incredible work! Thank you for doing this.

@Yuripetusko, can I tease you with that idea to see what of those tools we can start "Atomically" move from BOAT into OnchainKit.

Yuripetusko:
> Incredible work! Thank you for doing this.
> 
> @Yuripetusko, can I tease you with that idea to see what of those tools we can start "Atomically" move from BOAT into OnchainKit.

Yes 100%, let me know what you have in mind. I must say that I am a bit swamped at work these weeks, but happy to do some bits incrementally in between.

Zizzamia:
@Yuripetusko would love to hear your thoughts with an open issue, and just brainstorm some ideas around IPFS and NFT utilities to have in OnchainKit.

Than, happy to take some or all the work if need it. I just want to take advantage of your developer experience in see what we could add in OnchainKit.

Yuripetusko:
> @Yuripetusko would love to hear your thoughts with an open issue, and just brainstorm some ideas around IPFS and NFT utilities to have in OnchainKit.
> 
> Than, happy to take some or all the work if need it. I just want to take advantage of your developer experience in see what we could add in OnchainKit.

Cool, I do have some thoughts on this. Been spending a lot of time in wagmi/viem discord and got an idea of what sort of abstraction on top of wagmi/viem people are missing. I will message you on warpcast later today or tomorrow

PR #440, opened 2024-02-25
Improve ipfs util

### LINKED ISSUE #441: Feature Request: Improve ipfs utils
### Describe the solution you'd like

Currently ipfs utils consist of a single `ipfsToHTTP` util that assumes that uri is `ipfs://{CID}` format. However we have been running an NFT marketplace for the last 3 years and seend all sorts of stuff being put in there. Http links with gateway and cid, http links without cid, if http link then http:// instead of https://, base64 strings, uris with directory/path reference, etc. 
So it would be nice for ipfs util to handle wider range of cases

### Describe alternatives you've considered.

I created a PR (will link in a sec), that uses [`is-ipfs`](https://github.com/ipfs-shipyard/is-ipfs) library to check that CID is present in a string, and then re-construct a url based on that with provided gateway or default gateway

Yuripetusko:
**What changed? Why?**
- Use `is-ipfs` library to validate IPFS CID in provided uri/url
- Handle IPFS URIs with folder/directory path structure
- Handle HTTP IPFS URLs (many contracts use direct gateway urls in their contracts). And convert them to provided gateway
- Use provided gateway on all required fields (previously metadata fetching was not using gateway)
- Add more default gateway options, as ipfs.io fails often, so it is easier to test/develop before getting dedicated gateway

**Notes to reviewers**

**How has it been tested?**
`yarn test`

Fixes #441


template/web/next.config.js
```diff
@@ -21,6 +21,18 @@ const nextConfig = withPWA({
...
+        protocol: 'https',
+        hostname: 'cloudflare-ipfs.com',
+      },
+      {
+        protocol: 'https',
+        hostname: 'gateway.pinata.cloud',
+      },
+      {
+        protocol: 'https',
+        hostname: 'nftstorage.link',
```
Zizzamia, line 34: Oh that's dope. Teach me a bit, how do they work together?

template/web/src/utils/ipfs.ts
```diff
@@ -1,18 +1,128 @@
+import { cid } from 'is-ipfs';
```
Zizzamia, line 1: General thinking, I like a lot what's going on here.

Why some of these utility don't live in `is-ipfs` or, do we see other libraries have some of those?

Or should any of those live inside OnchainKit?

Zizzamia: @Yuripetusko thoughts on this?

Yuripetusko: [`is-ipfs`](https://github.com/ipfs-shipyard/is-ipfs) is great, but it mostly just checks if url has ipfs, it doesn't have utils to replace gateway or add gateway and preserve the path etc. But regarding onchainkit I agree, I can create a similar PR there if this makes sense.

[I actually have all of this code in a library already](https://www.npmjs.com/package/@rmrk-team/ipfs-utils). Just need to add docs. Perhaps instead of having is-ipfs and all this code explicitly, our library will work? Otherwise I can just duplicate some of it in Onchainkit

template/web/next.config.js
Yuripetusko: This is mostly telling nextjs to allow to load images from these domains in nextjs NextImage component

template/web/src/utils/ipfs.ts
Zizzamia: Let's probably move part of it inside OnchainKit.

We can proabably create a new section called IPFS.

Yuripetusko: Ok will do tomorrow. I'll close this PR then and later will create a new one where onchainkit utils are used

danimesq:
Hey @Yuripetusko 👋
Why you closed the PR?

PR #480, opened 2024-05-07
Fix setup error because of mistakes in file names

### LINKED ISSUE #479: Bug: Setup error - incorrect file renaming
### Describe the bug and the steps to reproduce it

When creating app with `npx @coinbase/build-onchain-apps@latest create`, the process will contain an error: `Error while renaming directories: Error: ENOENT: no such file or directory, rename '/home/.../web/src/OnchainProvidersWithRK.ts' -> '/home/.../web/src/onchainProviders.ts'`
This is because the existing file is `.tsx`, not `.ts`. Furthermore, capitalization is also incorrect (should be `OnchainProviders`, not `onchainProviders`).

### What's the expected behavior?

No errors

### What version of the libraries are you using?

latest

matejos:
**What changed? Why?**
Fixes #479

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/create/setupProject.ts
+++ b/src/create/setupProject.ts
@@ -144,16 +144,16 @@ export async function setupProject(projectDir: string, project) {
 
     if (project.pickSmartWallet) {
       removeDownloadedApps(projectDir + '/web/src/store/createWagmiConfigWithRK.ts');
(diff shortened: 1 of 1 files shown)


Zizzamia (approved):
Great catch!

Thank you @matejos for this fix

PR #486, opened 2024-05-10
feat: use api route for CDP RPC URL

0xAlec:
**What changed? Why?**
- hide API key for live app behind api route

**Notes to reviewers**
we will need to update the Vercel envars before we deploy this PR to production 
`NEXT_PRIVATE_RPC_URL` and `NEXT_PRIVATE_PAYMASTER_URL`

**How has it been tested?**
locally


arsood (approved):

template/web/src/store/createWagmiConfig.ts
```diff
@@ -8,6 +8,7 @@ export function createWagmiConfig(rpcUrl: string, projectId?: string) {
     console.log('projectId:', projectId);
   }
 
+  // Temporary hack, until we configure a FE page in OnchainKit to copy just the API key
```
Zizzamia, line 11: @0xAlec @cpcramer can we create an Issue on this, so we don't lose track of it.

template/web/app/api/rpc/route.ts
```diff
@@ -0,0 +1,30 @@
+import { NextRequest, NextResponse } from 'next/server';
+
+const rpcUrl = process.env.NEXT_PRIVATE_RPC_URL;
+
+export async function POST(req: NextRequest) {
```
Zizzamia, line 5: @0xAlec @cpcramer can we add some docs that explain more what's going on here.

Remember, many junior will use BOAT without clear context of the internal. So it's important to be as verbose as possible.

antoncoding:
My feedback as a template user: I think this change is unnecessary. Routing the api call to nextjs backend adds unnecessary overhead, most RPC providers already assume this project API keys are public, so they use other ways like "allowed domain" to restrict access and who can use those keys. Making it behind the proxy means those settings will become ineffective

PR #510, opened 2024-05-31
fix: Update Paymaster URL environment variable. 

cpcramer:
**What changed? Why?**
Update PaymasterURL from `NEXT_PRIVATE_PAYMASTER_URL` to `NEXT_PUBLIC_PAYMASTER_URL`.
- There were two issues here. First, we are accessing this environment variable on the client side, in Nextjs applications environment variables with the `NEXT_PRIVATE` prefix are not available. Therefore we were returning `undefined`. The second issue was that our production paymaster url was incorrect, and was actually a Base paymaster url rather than the intended CDP paymaster url. 

- Updated the `transactionID` to `callID`. This is a different ID that we were mistakingly calling the transaction id. It's the call id, which is what CallStatus uses the call id to fetch and display the status of a writeContracts call.

Tested locally and in dev:
Screenshot 2024-06-01 at 1 57 26 PM

Confirmed that we are using the CDP paymaster:
Screenshot 2024-06-01 at 1 58 28 PM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/template/web/app/paymaster-bundler/_components/PaymasterBundlerDemo.tsx
+++ b/template/web/app/paymaster-bundler/_components/PaymasterBundlerDemo.tsx
@@ -8,13 +8,13 @@ import { CallStatus } from './CallStatus';
 
 // Use the local API URL to target the Paymaster directly without a proxy
 // if running on localhost, otherwise use the Paymaster Proxy.
-const paymasterURL = process.env.NEXT_PRIVATE_PAYMASTER_URL;
+const paymasterURL = process.env.NEXT_PUBLIC_PAYMASTER_URL;
 const isLocalEnv = isLocal();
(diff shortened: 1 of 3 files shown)


template/web/app/paymaster-bundler/_components/PaymasterBundlerDemo.tsx
```diff
@@ -76,12 +76,7 @@ export default function PaymasterBundlerDemo() {
         >
           Mint NFT
         </button>
-        {transactionID && <CallStatus id={transactionID} />}
-        {transactionID && (
-          <div className={clsx('mt-4 overflow-x-auto rounded-lg bg-gray-800 p-4 text-lg')}>
-            <strong>Transaction ID:</strong> <span className="break-all">{transactionID}</span>
-          </div>
-        )}
```
cpcramer, line 84: With this fix, the user has no need to see the call id.

Zizzamia (approved):

PR #513, opened 2024-06-07
fix: Paymaster Proxy URL

cpcramer:
**What changed? Why?**

We are checking if process.env.NEXT_PUBLIC_VERCEL_ENV is undefined which means that we are on local. Therefore, we will hit our Paymaster directly. 

When deployed we are using the [System Environment Variable](https://vercel.com/docs/projects/environment-variables/system-environment-variables) `NEXT_PUBLIC_VERCEL_PROJECT_PRODUCTION_URL`.
**Notes to reviewers**

**How has it been tested?**
Tested locally and in Vercel Dev.


0xAlec (approved):

template/web/app/paymaster-bundler/_components/PaymasterBundlerDemo.tsx
```diff
@@ -2,15 +2,15 @@ import clsx from 'clsx';
...
 import { CallStatus } from './CallStatus';
 
-// Use the local API URL to target the Paymaster directly without a proxy
-// if running on localhost, otherwise use the Paymaster Proxy.
-const paymasterURL = process.env.NEXT_PUBLIC_PAYMASTER_URL;
-const isLocalEnv = isLocal();
-const defaultUrl = isLocalEnv ? paymasterURL : `${document.location.origin}/api/paymaster-proxy`;
+// Target the Paymaster directly without a proxy if running on localhost.
+// Use the Paymaster Proxy when deployed.
+const isLocalEnv = process.env.NEXT_PUBLIC_VERCEL_ENV === undefined;
```
Zizzamia, line 10: What's wrong with `isLocal`? should we fix that?

cpcramer: `isLocal` wasn't working for me. It was always returning that the application was running locally, even when deployed.

cpcramer: I think that we can just take advantage of the Vercel [System Environment Variable's](https://vercel.com/docs/projects/environment-variables/system-environment-variables)

cpcramer: It actually looked like the issue was with [getCurrentEnvironment](https://github.com/coinbase/build-onchain-apps/blob/95085424a550931d5d9c5349d070aa47b4cbdd9e/template/web/src/store/environment.ts#L18) - a function that `isLocal` relies on.

Zizzamia: My advice here is, to make sure `isLocal()` works well. And just use that everywhere.

And avoid doing side solution.

In a few words, improve what we have, and avoid create N+1 solution for the same problem.

cpcramer: Makes sense! I updated getCurrentEnvironment to have the NEXT_PUBLIC prefix and it now works. I've tested in local and dev!

Zizzamia (approved):

PR #525, opened 2024-07-17
fix #517

### LINKED ISSUE #517: Bug: Create Accepts `.` as the name of the project but then errors
### Describe the bug and the steps to reproduce it

1. Create a folder called `my-test`
2. In a console in `my-test` run `npx @coinbase/build-onchain-apps@latest create`
3. Enter `.` as the name of the project
4. Select any options desired
5. Observe failure:  Directory /Users/briandoyle/src/Zora-Tutorial-Test already exists.

### What's the expected behavior?

Script should accept `.` as the project name and create a project in the root of the current folder 

### What version of the libraries are you using?

- BOAT v0.31.0

danimesq:
Hopefully this will fix the issue #517. THIS IS UNTESTED.


Zizzamia:
So we got rid of all the CLI part, and focusing more on making this template easier to fork.

